Issue #18, opened 2017-01-25
Clarify Decisions


robs16:
It is my understanding that decisions are not meant to direct the EMR to perform the decision action (create/delete). Instead they're meant to be a notification to the EMR after the action has already been executed (out of band) by the CDS vendor. 

The documentation needs to reflect this understanding, as it is easy to misunderstand and assume that decision should cause the EMR to perform the requested updates.

jmandel:
@robs16 your description here isn't quite right. The idea of a "decision" is to communicate a user's intent to the EHR — the user had a conversation with an external app and *decided* something (e.g. decided to change drugs while writing a prescription), but the thing *isn't done yet* (e.g. the changes hasn't been made).

The EHR's job upon learning about a "decisoin" is to update the UI to reflect this decision, allowing the user to continue in the workflow from where they left off. 

This is super-helpful in cases where EHR's don't support API driven writes. For example, the EHR probably doesn't support an API for creating a new prescription —so the CDS Hooks framework allows an alternative, where the service returns a "decision" that the EHR can then render within its native UX. Think of it as a bridge to writing

For data types where the external service *can* write (e.g. adding a new problem to the problem list), the service could just go ahead and issue the `POST` call. Then after returning to the EHR, the service might just include an "information card" saying (for example) "hypertension added to the problem list" (for example).

robs16:
@jmandel Thanks for the clarification. From what you said there appears to be (at least) two distinct ways for the CDS service to provide the decision output and it is dependent on the consumer of the CDS service. This implies a tight coupling between the consumer and provider (custom integration), which IMHO is not ideal for an interface standard. I would prefer the the decisions are communicated in the same manner in both of the scenarios, potentially with an indicator to show if a decision has already been performed.

jmandel:
Thanks for the comments @robs16! Can you clarify which "two distinct ways" you're referring to?

If you're talking about REST API calls vs CDS Hooks Decision responses... my apologies if I was unclear. These would be serving different use cases. One (REST API calls) is for changes that could be made directly to a system (by anyone authorized client, EHR it's a CDS service, SMART app, or otherwise); the second (decisions) is for changes that can't be made directly to an EHR's system, and so instructions they're merely proposed via the UI of the electronic health record, for a user to confirm/complete.

If this explanation still seems wrong, I wonder if you might give an example of the alternative approach that you're expecting. I really appreciate your attention here!

robs16:
[...]

PR #26, opened 2017-03-30
Decision activity proposal

bkaney:
This implements point 3 from https://github.com/cds-hooks/docs/issues/18#issuecomment-282141134

This adds a new concept to the decision return from CDS service to hint at a subsequent action and related resources. This also generalizes the name of "hook catalog" to be called "activity catalog", as it serves as describing actions, of which could be used as web hooks.


jmandel:
Thanks @bkaney! I think the key contributions here are:

1. Generalizing the activities catalog to work in both directions: an EHR can tell a service "hey, the user is doing this activity", and an app can tell the EHR "hey, here's a good activity for the user to do next". 

2. Updating the decision data model to support this communication back to the EHR

I'm definitely in favor.

I noticed, in the process, that our docs for `decisions.create` are missing something: they specify a list of ids, but what we actually need (and the way our test implementations work) is a list of resources. Does this match your expectations?

bkaney:
> I noticed, in the process, that our docs for decisions.create are missing something: they specify a list of ids, but what we actually need (and the way our test implementations work) is a list of resources. Does this match your expectations?

Yea, they could be resources. However, as we aren't the FHIR server persisting the resource, we'd have to figure out a way to reference them (in the `actions` structure). Perhaps the CDS service would mint an ID for each resource and use #anchor-tag style contained reference?

isaacvetter:
Hey Guys,

This is a really neat idea. I think that a decision returning an EHR activity is great functionality and we should try to make it happen.

1. I'm not sure that a hook and an activity are actually the same thing. For example, take a look at the [proposed hooks](https://github.com/cds-hooks/docs/wiki/Proposed-Hooks), these seem reasonable as hooks, but aren't specific enough to identify an activity.

2. How does the decision define the activity's "parameters". I think that your example specifies a CarePlan as input to a medication-prescribe activity. What does that mean? Is there a medication that should be pre-populated into the EHR's medication-prescribe activity?

Isaac

bkaney:
Hi @isaacvetter --

> I'm not sure that a hook and an activity are actually the same thing. For example, take a look at the proposed hooks, these seem reasonable as hooks, but aren't specific enough to identify an activity.

I see what you mean. I was thinking we avoid having both a "hook catalog" and then another "post hook catalog". Especially after observing the contents both describe what I think of as activities. Instead, there would be a single catalog where the usage context would be as a "hook" or "next action".

> How does the decision define the activity's "parameters". I think that your example specifies a CarePlan as input to a medication-prescribe activity. What does that mean? Is there a medication that should be pre-populated into the EHR's medication-prescribe activity?

I updated the example to be a bit more simplified, including expanding the resources inline. Have a look and see if this addresses your questions.

kpshek:
Given that we've moved decisions out of 1.0, and considering that our specification has changed quite radically (eg, slate -> mkdocs) since this PR was opened, I'm closing this.

When we tackle decisions post 1.0, we can look at this PR for great ideas for possible implementation strategies.

PR #27, opened 2017-03-30
Remove `redirect`, add `postMessage` proposal

bkaney:
This is a proposal for the improvement discussed in the Google group regarding using postMessage instead of a full redirect to inform the EHR when "done" using an app.

see: https://groups.google.com/forum/#!topic/cds-hooks/5Vxwoa2FLlo

### DIFF
--- a/source/includes/_cds_services.md
+++ b/source/includes/_cds_services.md
@@ -81,7 +81,6 @@ curl
    "hookInstance" : "d1577c69-dfbe-44ad-ba6d-3e05e953b2ea",
    "fhirServer" : "http://hooks.smarthealthit.org:9080",
    "hook" : "patient-view",
-   "redirect" : "http://hooks2.smarthealthit.org/service-done.html",
    "user" : "Practitioner/example",
    "context" : [],
    "patient" : "1288992",
@@ -115,7 +114,6 @@ Field | Description
 <nobr>`hookInstance`</nobr> |*string*.  A UUID for this particular hook call (see more information below)
 `fhirServer` |*URL*.  The base URL EHR's [FHIR](https://www.hl7.org/fhir/) server. The scheme should be `https`
 `oauth` | *object*. The OAuth2 authorization providing access to the EHR's FHIR server (see more information below)
-`redirect` |*URL*.  The URL an app link card should redirect to (see more information below)
 `user` |*string*.  The FHIR resource type + id representing the current user.<br />The type is one of: [Practitioner](https://www.hl7.org/fhir/practitioner.html), [Patient](https://www.hl7.org/fhir/patient.html), or [RelatedPerson](https://www.hl7.org/fhir/relatedperson.html).<br />For example, `Practitioner/123`
 `patient` |*string*.  The FHIR `Patient.id` of the current patient in context
 `encounter` |*string*.  The FHIR `Encounter.id` of the current encounter in context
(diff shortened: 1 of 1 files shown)


isaacvetter:
Hi Brian,

What would this mean for apps that aren't hosted in an iframe?

Isaac

bkaney:
> Hi Brian,
> 
> What would this mean for apps that aren't hosted in an iframe?
> 
> Isaac

Hi Isaac -- `postMessage` enables cross-origin communication. This can mean iframe/parent, but also can mean when a window opens another window.

jmandel:
Thanks @bkaney! One aspect we should cover here (well, in any discussion of `postMessage`) is the target origin: we should specify how to lock this down on both sides, so that the CDS service's web UI (SMART app) can post messages _to_ a known target domain, and the EHR's listener window can filter by messages coming _from_ a known target domain. Thoughts on how to tie this back into our discovery/launch protocol?

bkaney:
> Thanks @bkaney! One aspect we should cover here (well, in any discussion of postMessage) is the target origin: we should specify how to lock this down on both sides, so that the CDS service's web UI (SMART app) can post messages to a known target domain, and the EHR's listener window can filter by messages coming from a known target domain. Thoughts on how to tie this back into our discovery/launch protocol?

Hmm, good point. I wonder if there is a way to use the existing HookInstance identifier somehow? I'll think about this over the next day or so and update the PR.

bdoolittle:
Hi All,

postMessage seems like a nice alternative to the redirect url, we're really excited about its ability to enable a more responsive user experience. My team has a few points we'd like to add for consideration. 

First, we agree with Josh that specifying targetOrigin when sending the message and checking the origin on the receiving side is a good security measure to add into the spec. Should the SMART App also be sending a security token (or other mechanism) to add an additional level of authentication?

Second, a web-based EHR may want to open the SMART App in a new tab rather than an iframe, this introduces the possibility that the user may exit the EHR before completing the SMART App interaction. In this case, there would be no window to send a postMessage to. Would the CDS Service know that the postMessage was not received? Should there be some acknowledgement from the EHR to the SMART App indicating that the EHR successfully received the postMessage? This way, the SMART App could indicate to the user that there was an error in notifying the EHR about the decision.

Third, there may be implementations where the redirect URL solution is more accessible than postMessage. Would it be reasonable for the hooks standard to be flexible and support both means of communication? The worry here is that the cds-hooks implementation becomes more complex.

jmandel:
I had a chance to talk with @bkaney in person yesterday  at a Boston area SMART  meetup. We discussed the key use  case here, which I summarize as "give apps a way to return control to the EHR, indicating what new orders or other changes the user has selected in-app, and possibly giving the EHR a hint about where to navigate next".

While postMessage definitely provides nice features (the possibility of two way communication, for example), I think we established that the current redirect-based approach *can* handle the use cases driving @bkaney's proposal here. The "trick" to making this work is that the EHR species the redirect URL, and the EHR can implement a static page (or even server side logic) at that URL to communicate back with the parent frame, previous tab, or wherever the EHR is running. In fact, this is how our test harness works today.

jmandel:
Note that I'm not saying we should reject this PR! I just want to make sure we're thinking through the tradeoffs.

bkaney:
I agree with @jmandel. There is two more things I was thinking about. 

1) I was thinking the `hookInstance` should be passed (as a query parameter, for instance) to the redirect URL. This would be helpful -or- necessary information for EHR to deal with navigation post-interaction.

2) I appreciate the "lowest common denominator" idea around a redirect URL. However, given a client-server EHR that is using some sort web dialog widget, having a javascript way to do this (e.g. `postMessage`) might be in fact the LCD...

wyacteen:
I'm a little late to this discussion but wanted to share some results of some quick prototyping I did to check browser compatibility.  I found that IE 10 and 11 were not able to use `postMessage` when when they were in separate tabs or windows. It only worked when iframed.

I also tested this on the current versions of Chrome, Edge, and Safari. Using `postMessage `worked in both iframed and new tab/window tests.

Has anyone else done similar tests?  I'm hoping I'm wrong about IE 10 & 11.

jmandel:
IE 10 has the challenges you've described.
https://stackoverflow.com/questions/16226924/is-cross-origin-postmessage-broken-in-ie10
covers the crux of it.

wyacteen:
In light of the limitations of using _postMessage_ with IE 10 & 11 that would require the SMART app to be iframed, is this cause for concern?  I'm not sure how much longer those browsers will be officially supported, but I know it can be difficult to get users to change their browsers, particularly when they are part of large organizations with IT departments that are in control of what's on their machines.  The nice thing about the redirect url is that it's browser agnostic, though does require a server.

kpshek:
Given that we've moved decisions out of 1.0, and considering that our specification has changed quite radically (eg, slate -> mkdocs) since this PR was opened, I'm closing this.

When we tackle decisions post 1.0, we can look at this PR for great ideas for possible implementation strategies.

Issue #47, opened 2017-05-08
Constrain prefetch to align with FHIR's DataRequirements


kpshek:
In discussion with @brynrhodes and @isaacvetter regarding aligning Clinical Reasoning and CDS Hooks, one of the topics of discussion was around how both projects handle the concept of prefetching data.

Prefetching aims to address the following:
1. Allows the EHR to satisfy data that it may have already (eg, in memory)
2. Allows the EHR to calculate an optimal set of queries necessary to satisfy data across multiple CDS Services

In Clinical Reasoning, this concept is defined in the [ServiceDefinition.dataRequirement](https://www.hl7.org/fhir/servicedefinition-definitions.html#ServiceDefinition.dataRequirement) using the [DataRequirement](https://www.hl7.org/fhir/metadatatypes.html#DataRequirement) type. Compared to our prefetch definition, DataRequirement is a much more constrained model. Based upon @brynrhodes implementation experience, the constraints of DataRequirements allow for the calculation of an optimal set of FHIR queries to satisfy multiple DataRequirements. Our current prefetch definition (any FHIR query) would not actually allow EHRs to achieve our 2nd goal above.

To help with this discussion, it would be good to take an inventory on the current prefetch FHIR queries for a variety of CDS Services that have been prototyped at Connectathons. With their feedback, we can see if constraining prefetch to the functionality of DataRequirements would still meet their needs.

jmandel:
@kpshek can you explain a bit about what you mean by "calculation of an optimal set of FHIR queries"? Is this about generating raw/internal database queries to retrieve the data you need? In general, the more capabilities you allow the harder this gets; but for simpler search strings (e.g. `Observation?patient={{Patient.id}}` you could always optimize these directly (by recognizing them explicitly as having some structure that you know how to handle well).

Obviously this is more work on the EHR side, but I'd imagine a couple of common cases would handle a good fraction of the load, while still giving service developers the flexibility to pre-fetch on more expressive REST API when needed. And if you see some pre-fetch requests that are "too hard" (whatever that might mean), you can always leave them out and allow the client to fetch directly.

All that said: I agree that it would be good to take stock of what queries people are actually running. On my list so far are:

 * patient (read)
 * search for most recent set of bilirubin labs for a specific patient
 * search for all med orders + statements for a patient
 * search for most recent height + weight for a patient
 * search for all allergies for a patient
 * practitioner (read)

[...]

PR #58, opened 2017-07-19
Constrain prefetch semantics and functionality to align with HL7 Clinical Reasoning

brynrhodes:
Adds restrictions to the allowed prefetch queries to reduce EHR implementer burden and align with the DataRequirement of FHIR Clinical Reasoning.

This pull request addresses issue #47.


kpshek:
@brynrhodes - Can you clarify the restrictions further? For example:

- What does "instance level read interactions" mean? I'm guessing this [read](https://www.hl7.org/fhir/http.html#read) in FHIR? If my understand of that is correct, what does the further qualification of "for the Patient resource only" mean?

- Does "type level search interactions" only mean `search` and not `create` or `history`? Why is `GET` called out? The EHR can satisfy the prefetch data however it sees fit and I'm not sure what GET vs POST would mean

Are these similar restrictions defined similar to how you have documented then here somewhere in Clinical Reasoning? If so, where?

My questions here are just an attempt to ensure that I and everyone else understands very clearly what we're proposing is & isn't allowed in prefetch queries. 😄

grahamegrieve:
I don't think these restrictions should be global. There might not even be a patient in context. More generally, what is most effective varies widely by context. There might not be the possibility for the cds-hooks server to pull further information from the EHR at all, for instance. 

I'm not opposed to being able to make - or even recommend - these restrictions (though they need more clarity as Kevin says) but they can't be applied always

brynrhodes:
@kpshek, yes, I'm thinking of the "read" and "search" interactions. I was trying to describe them in terms of the HTTP level FHIR API only, so that was the reason for the reference to GET, but after reading your updates to the prefetch on the other pull, I agree, and I've removed the references to HTTP verbs.

I've also expanded the Patient request to include resource for which an id is known (so it could include Practitioner or Encounter, for example).

As far as documentation about how these restrictions are used within Clinical Reasoning, there is documentation on the [DataRequirement](http://hl7.org/fhir/metadatatypes.html#datarequirement), are you thinking a link there would be helpful?

@grahamegrieve, agreed, I've relaxed the language to make them 'recommended', rather than 'required'.

PR #65, opened 2017-07-26
Add CORS documentation (fixes #1)

### LINKED ISSUE #1: Add CORS documentation
Several CDS service implementations have gotten stuck on CORS. We should add documentation on CORS and that services should implement support for this

kpshek:


isaacvetter:
Assuming the idea is that we'll also add OAuth2 documentation within this same security section - I like it.

kpshek:
> Assuming the idea is that we'll also add OAuth2 documentation within this same security section - I like it.

Exactly 😄

PR #70, opened 2017-07-26
Update the Ruby versions that Travis-CI builds use

kpshek:
Update the Ruby versions that Travis-CI builds use to avoid errors during builds for missing Rubies. Additionally, by specifying just the major and minor version and not the revision (eg, 2.3 instead of 2.3.0), this will allow Travis to use the latest revision/patch during builds.

This should fix the [recent Travis-CI build failures](https://travis-ci.org/cds-hooks/docs/jobs/257608496):

```
5.72s$ rvm use 2.1.0 --install --binary --fuzzy
ruby-2.1.0 is not installed - installing.
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm ruby-2.1.0 do rvm gemset create ' first, or append '--create'.
The command "rvm use 2.1.0 --install --binary --fuzzy" failed and exited with 2 during .
Your build has been stopped.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: false
 language: ruby
 
 rvm:
-  - 2.0.0
-  - 2.1.0
+  - 2.3
+  - 2.4
 
 cache: bundler
 script: bundle exec middleman build


kpshek:
For some reason, the Travis CI builds started failing as the Travis build machines could not find a precompiled Ruby 2.1.0. In researching this error, I found that the [best practice](https://github.com/travis-ci/travis-ci/issues/2220) is to reference Ruby versions up to the minor release (eg, 2.1 instead of 2.1.0) and let Travis choose the latest Ruby on that minor release.

Previously, we were building with both 2.0.0 and 2.1.0. This pull request changes that to simply 2.3 (so 2.3.x). I tried also building with 2.4 (so 2.4.x) but version of the json gem we're pulling in doesn't support Ruby 2.4 (see https://github.com/flori/json/issues/303).

I'm going to merge these changes into master and sync all branches so that the Travis CI builds start passing again.

PR #72, opened 2017-07-27
Define security model

kpshek:
This is an in-progress branch -- please don't bother reviewing right now. The purpose for issuing this pull request early is to get Heroku review apps deployed.

### DIFF
--- a/source/includes/_security.md
+++ b/source/includes/_security.md
@@ -1,5 +1,61 @@
 # Security
 
+CDS Hooks defines the agreed upon security model between an EHR and the CDS Service. Like SMART on FHIR, the security model of CDS Hooks leverages the same open and well supported standards like OAuth 2 and JSON web tokens. However, as CDS Hooks differs from SMART, the manny in which these standards are used is specific to CDS Hooks.
+
+## Trusting CDS Services
+
+As the EHR initiates every interaction with the CDS Service, it is responsible for establishing trust with the CDS Services it intends to call. This trust is established via a TLS connection to the CDS Service. Thus, all CDS Service endpoints must be deployed to a TLS protected URL (https). This includes both the Discovery and individual CDS Service endpoints.
+
+EHRs should use accepted best practices for verifying the authenticity and trust of these TLS connections. For instance, [rfc5280](https://tools.ietf.org/html/rfc5280) and [rfc6125](https://tools.ietf.org/html/rfc6125). Additionally, it is assumed that EHRs configure the CDS Services they connect to via some offline process according to the business rules and practices of both the EHR and CDS Service organizations.
+
+## Trusting EHRs
+
+Since the CDS Service is invoked by the EHR, the CDS Service does not have the same mechanism as EHRs to establish trust of the EHR invoking it. To establish trust of the EHR, [JSON web tokens](https://jwt.io/) are used.
+
+Each time the EHR makes a request to the CDS Service, it should send an `Authorization` header where the value is `Bearer <token>`, replacing `<token>` with the actual JWT. Note that this is for every single CDS Service call, whether that be Discovery calls, CDS Service invocations, etc.
+
+> Example JSON web token payload
+
+```json
+{
+  "iss": "https://fhir-ehr.example.com/",
+  "aud": "https://cds.example.org/cds-services",
+  "exp": 1422568860,
+  "jti": "2af5e8a5-b5ef-4fd0-8d93-c560dac40c98"
+}
+```
+
+> Using the above JWT payload, the complete JWT as passed in the Authorization HTTP header would be:
+
+```
+Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwczovL2ZoaXItZWhyLmV4YW1wbGUuY29tLyIsImF1ZCI6Imh0dHBzOi8vY2RzLmV4YW1wbGUub3JnL2Nkcy1zZXJ2aWNlcyIsImV4cCI6MTQyMjU2ODg2MCwianRpIjoiMmFmNWU4YTUtYjVlZi00ZmQwLThkOTMtYzU2MGRhYzQwYzk4In0.ahR57rtcMFhvrHEEo9w13vVdLrhZs_gRY2NV6R2GAoU
+```
+
+The JWT from the EHR is signed with the EHR's private key and contains the following fields:
+
+Field | Value
+----- | -----
+iss | The base URL of the EHR's FHIR server. This must be the same URL as the `fhirServer` field in a CDS Service request.
+aud | The URL being invoked by the EHR.
+exp | Expiration time integer for this authentication JWT, expressed in seconds since the "Epoch" (1970-01-01T00:00:00Z UTC). This time MUST be no more than five minutes in the future.
+jti | A nonce string value that uniquely identifies this authentication JWT.
+
+Note that the use of JWT in CDS Hooks resembles the [SMART on FHIR Backend Services](http://docs.smarthealthit.org/authorization/backend-services/) usage of JWT. However, there are differences due to the differing use cases and concerns between CDS Hooks and SMART Backend Services.
+
+[https://jwt.io/](https://jwt.io/) is a great resource not only for learning about JSON web tokens, but also for parsing a JWT value into its distinct parts to see how it is constructed. Try taking the example JWT here and pasting it into the form at [https://jwt.io/](https://jwt.io/) to see how the token is constructed.
+
+<aside class="notice">
+TODO: Need to propose how the EHR public key is discovered. Preference would be a further extension in the EHR FHIR server Conformance resource.
+</aside>
+
+### Mutual TLS
+
+[Mutual TLS](https://en.wikipedia.org/wiki/Mutual_authentication) (mTLS) can be used between an EHR and CDS Service and that would allow the CDS Service to establish trust of the EHR. However, if mTLS is used, this should be in addition to using JSON web tokens to establish trust of the EHR. As mTLS is not well supported across all platforms and technologies, it is not the standard means of establishing trust with the EHR.
+
+## FHIR Resource Access
+
+TODO
+
 ## Cross-Origin Resource Sharing
 
 [Cross-origin resource sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS) is web security mechanism that is built into browsers. In short, CORS allows servers to control how browsers access resources on the server, including the accessible HTTP response headers. CORS is only honored by web browsers and as such, is a client-side security mechanism.


source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+
+### Obtaining an Access Token
+
+In SMART on FHIR, the SMART app requests and ultimately obtains an access token from the Authorization server using the SMART launch workflow. This process utilizes the authorization code grant model as defined by the OAuth 2.0 Authorization Framework in [rfc6749](https://tools.ietf.org/html/rfc6749).
+
+With CDS Hooks, the EHR provides the access token directly in the request to the CDS Service. Thus, the CDS Service does not need to request the token from the authorization server as a SMART app would. This is done purely for performance reasons as the authorization code grant model in OAuth 2 involves several HTTPS calls and redirects. In contrast with a SMART app, a CDS Service may be invoked many times during a workflow. Going through the authorization code grant model on every hook invocation would likely result in a slow performing CDS Service due to the authorization overhead.
```
olbrich: > In contrast with a SMART app

In contrast to a SMART app

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+The access token is specified in the CDS Service request via the `fhirAuthorization` request parameter. This parameter is an object that contains both the access token as well as other related information.
+
+Field | Description
+----- | -----------
+`access_token` |*string*. This is the OAuth 2 access token that provides access to the FHIR server.
+`token_type`   |*string*. Always the value `code`
+`expires_in`   |*integer*. The lifetime in seconds of the access token.
+`scope`        |*string*. The scopes the access token grants the CDS Service.
+
+It is recommended that the `expires_in` value be very short lived as the access token should be treated as a transient value by the CDS Service.
```
olbrich: What timestamp is the `expires_in` value being used with to determine if it has expired?

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+
+It is recommended that the `scope` value contain just the scopes that the CDS Service needs for its logic and no more.
+
+As the CDS Service is executing on behalf of a user, it is important that the data the CDS Service has access to is under the same restrictions/authorization as the current user. As such, the access token should be scoped to:
+
+- The CDS Service being invoked
+- The current user
+- The current patient
```
olbrich: Many EHRs are multi-tenant and may need the token to be scoped by a practice/organization.  Some organization models also allow users to have access to multiple organizations so it is not safe to assume that the current user maps one-to-one with an organization or practice.

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+
+### Frequently Asked Questions
+
+1. How is the user (clinician) prompted to authorize the CDS Service to obtain an access token on behalf of the user? If we're aligning with how security works for SMART apps, this is what happens today.
+
+Not quite. There is actually no requirement that there be any user interaction to authorize a SMART app to obtain an access token. More generally, the OAuth 2 framework does not prescribe *how* authorization is determined. That is, authorization may be determined via some user interruptive decision, some pre-existing business rules, or even by random chance! The authorization server is free to determine how it grants access to the resources it protects.
+
+With SMART on FHIR, we have seen real production behavior in which the authorization server:
+
+- Grants access to practictioner facing SMART apps via some predefined business arrangement that was done out of bounds. The user (practitioner) never is asked to authorize the SMART app as their organization (hospital) has already made this decision for them.
```
olbrich: > out of bounds

out-of-band

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+- Grants access to patient facing SMART apps by asking the user explicitly for permission to both launch the SMART app as well as what specific scopes (data permissions) the SMART app may havel
```
olbrich: > havel

have

kpshek: Thanks! Fixed in dde069c

source/includes/_security.md
kpshek: Thanks! Fixed in dde069c

source/includes/_security.md
kpshek: Thanks! Fixed in dde069c

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+<aside class="notice">
+TODO: Need to propose how the EHR public key is discovered. Preference would be a further extension in the EHR FHIR server Conformance resource.
+</aside>
+
+<aside class="notice">
+TODO: Need to outline the risks of noalg signatures
+</aside>
+
+<aside class="notice">
+TODO: Need to exposing client_id and scopes in Discovery
```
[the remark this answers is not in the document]
kpshek: Thanks! Fixed in dde069c

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
 # Security
 
+<aside class="notice">
+The proposed security model has not yet received implementer feedback and as such, is subject to change. Maintaining both a secure model and an easy implementation experience is a key concern of our impending 1.0 release. As such, we encourage open feedback on this proposed approach on <a href="https://github.com/cds-hooks/docs/issues/7">this Github issue</a> from all of our stakeholders.
+</aside>
+
+CDS Hooks defines the agreed upon security model between an EHR and the CDS Service. Like SMART on FHIR, the security model of CDS Hooks leverages the same open and well supported standards like OAuth 2 and JSON web tokens. However, as CDS Hooks differs from SMART, the manny in which these standards are used is specific to CDS Hooks.
```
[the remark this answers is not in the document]
kpshek: Thanks! Fixed in dde069c

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+aud | The OAuth 2 client id of the CDS Service
+exp | Expiration time integer for this authentication JWT, expressed in seconds since the "Epoch" (1970-01-01T00:00:00Z UTC).
+iat | The time at which this JWT was issued, expressed in seconds since the "Epoch" (1970-01-01T00:00:00Z UTC).
+
+Note that the use of JWT in CDS Hooks resembles the [SMART on FHIR Backend Services](http://docs.smarthealthit.org/authorization/backend-services/) usage of JWT. However, there are differences due to the differing use cases and concerns between CDS Hooks and SMART Backend Services.
+
+[https://jwt.io/](https://jwt.io/) is a great resource not only for learning about JSON web tokens, but also for parsing a JWT value into its distinct parts to see how it is constructed. Try taking the example JWT here and pasting it into the form at [https://jwt.io/](https://jwt.io/) to see how the token is constructed.
+
+<aside class="notice">
+TODO: Need to propose how the EHR public key is discovered. Preference would be a further extension in the EHR FHIR server Conformance resource.
```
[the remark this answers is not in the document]
kpshek: Yes, just like SMART, CDS Hooks is going to have to allow for both DSTU2, STU3, and greater FHIR usage. Conformance/Capability Statement is one example, MedicationOrder/MedicationRequest is another, etc.

source/includes/_security.md
kpshek: Normally, the developer (client role in OAuth 2 speak) would use this value to compare against the current date/time and know when they need to request a refresh token. (See [RFC 6749 section 4.2.2](https://tools.ietf.org/html/rfc6749#section-4.2.2) for more info on this field).

With CDS Hooks, since the usage of the FHIR access token is transient and the CDS Service is running for no more than a couple of seconds, the CDS Service developer shouldn't need to worry about this. I included it here for completeness as it is part of the access token response from the authorization server.

I'll add further documentation around this to make this more clear.

source/includes/_security.md
kpshek: I'm not assuming the user maps one-to-one with and org or practice. The point I was trying to make here is that the access token needs to be narrowly scoped such that if it it cannot be misused by the CDS Service to obtain data the current user cannot see. This line echoes that point:

> As the CDS Service is executing on behalf of a user, it is important that the data the CDS Service has access to is under the same restrictions/authorization as the current user.

Is there a way you can suggest that I make this more clear?

(As an aside, Cerner's EHR can be run in a single or multi-tenant mode so I am always thinking of both models to make sure whatever we do works for both)

kpshek:
Thanks for the great suggestions & feedback @olbrich - I've replied to each inline above

source/includes/_security.md
olbrich: It might be necessary to optionally pass additional context (i.e, add an `organization_id`) to clearly identify which organization owns the data.

kpshek: Do you have a SMART app today? If so, how are you handling this today?

source/includes/_security.md
```diff
@@ -1,5 +1,132 @@
...
+
+In SMART on FHIR, the SMART app requests and ultimately obtains an access token from the Authorization server using the SMART launch workflow. This process utilizes the authorization code grant model as defined by the OAuth 2.0 Authorization Framework in [rfc6749](https://tools.ietf.org/html/rfc6749).
+
+With CDS Hooks, the EHR provides the access token directly in the request to the CDS Service. Thus, the CDS Service does not need to request the token from the authorization server as a SMART app would. This is done purely for performance reasons as the authorization code grant model in OAuth 2 involves several HTTPS calls and redirects. In contrast to a SMART app, a CDS Service may be invoked many times during a workflow. Going through the authorization code grant model on every hook invocation would likely result in a slow performing CDS Service due to the authorization overhead.
+
+```json
+{
+  "fhirAuthorization" : {
+    "access_token" : "some-opaque-fhir-access-token",
+    "token_type" : "code",
```
robs16: Can these fields be changed to be consistent with the "fhirAuthorization", "hookInstance", and other fields that do not use an underscore?

isaacvetter: Hey @robs16 , these properties names are intentionally the same as those defined by SMART and OAuth -- which do use underscores. 

What problem are the underscores causing for you?

Isaac

http://www.hl7.org/fhir/smart-app-launch/

robs16: it doesn't cause an issue - its just inconsistent. It sounds like there is a good reason for the inconsistency though, so I'll retract my objection.

source/includes/_security.md
olbrich: We haven't yet solved that problem, although it's going to become an issue for us soon.

kpshek:
At the San Diego Connectathon this weekend (2017-09-09), a large group of us participating in the CDS Hooks track discussed this issue in an offline discussion. This group of ~30 represented a broad set of stakeholders, from multiple EHR vendors, several CDS Service providers, healthcare organizations, and those representing various HL7 working groups. What follows is the summary of our discussion and the consensus of the group.

We affirmed our desire to define a standard approach to security within CDS Hooks. Additionally, our current proposed strategy has received overall positive implementer feedback. Particulars around the JWT did receive feedback that will warrant changes. Today, our JWT is based upon the JWT that SMART uses as part of the OpenID Connect workflow to assert a user identity. Based upon implementation feedback and considerations, we agreed that our use of a JWT is not to establish user identity but rather a bearer token for the CDS Service to authorize an EHR. As such, the content of our JWT would change, the details of which are below.

- `sub`: remove this field as our JWT is not trying to assert a user identity
- `iss`: Change this from the FHIR server URL to the URL of the actual issuer. This could still be the FHIR server but not guaranteed. For instance, in the case of the CDS Hooks Sandbox, the issuer would change to http://sandbox.cds-hooks.org. The reason for this change is that the `iss` value should be the entity responsible for signing the JWT. In the case of the CDS Hooks Sandbox, or other CDS Hooks systems which are capable of connecting to multiple/arbitrary FHIR servers, having the issuer be strictly the FHIR server is wrong.
- `aud`: Change this to the URL of the CDS Service being invoked. Previously, this was the OAuth 2 client id of the CDS Service but the consensus of the community was that the CDS Service does not need to know this value. Additionally, making the audience the URL of the CDS Service being invoked will protect against JWTs be sent to incorrect audiences.
- `jti`: Add this new field which is a nonce (UUID) to protect against replay attacks.

I'll be updating the documentation to reflect these changes shortly.

isaacvetter:
Kevin, 

I re-read the whole security section and came up with one substantive suggestion and many non-substantive suggestions.

== Substantive ==
This sentence:

> As such, the access token should be scoped to:
> The CDS Service being invoked
> The current user
> The current patient 

I don't think that the access_token should be limited to a specific patient. The CDS Service is acting on behalf of the user and therefore should have the same level of data access as the current user.  

==Non-Substantive ==

1. Replace "The proposed security model has not yet received implementer feedback and as such, is subject to change" with "The CDS Hooks security model is undergoing a rigorous security assessment and as such, may be subject to change"

1. The language in the security section uses conditional sentences, which weaken the specification and use more words than necessary. Notably, look at the use of  the word "should". Some of these uses should be replaced with current tense, factual statements, using words like "is" and "does".  Let's only use "should" when there really is a choice to be made by an implementer. 
* Each time the EHR makes a request to the CDS Service, it _should send_ an Authorization header 
* Each time the EHR makes a request to the CDS Service, it _sends_ an Authorization header 

* When the EHR invokes the CDS Service discovery endpoint, the aud value _should be either_
* When the EHR invokes the CDS Service discovery endpoint, the aud value _is either_



1. These two sentences are overly complicated: "Mutual TLS (mTLS) can be used between an EHR and CDS Service and that would allow the CDS Service to establish trust of the EHR. However, if mTLS is used, this should be in addition to using JSON web tokens to establish trust of the EHR." Replace with: 
"Mutual TLS (mTLS) _may be used alongside JSON web tokens to establish trust of the EHR by the CDS Service_. "

1. Using the slate doc column format (instead of the MkDocs inline code snippets), this sentence is pretty confusing: "Try taking the example JWT here and pasting it into the form at https://jwt.io/ to see how the token is constructed."

1. " TODO: Need to outline the risks of noalg signatures" Do we really? 

1. Suggest replacing " TODO: Need to propose how the EHR public key is discovered. Preference would be a further extension in the EHR FHIR server Conformance resource." with " TODO: CDS Hooks doesn't yet specify how a CDS Service discovers the public key that the EHR uses to sign the JWT. "

1. "likely result in a slow performing CDS Service due to the authorization overhead." should either use the term "slow" or "poorly performing" in place of "slow performing".

kpshek:
@isaacvetter & @brynrhodes - I've addressed all of your comments in 3cb827a.

Could you guys both take a look at this document and approve the merge if it looks good?

PR #73, opened 2017-07-31
Travis deploys

mattberther:
@kpshek This is untested, but should do the job. I've already added the encrypted deploy key to travis (and to the repo here). Take a look through, and if you like it, merge it.

Please let me know if you have any questions.


kpshek:
Thanks for putting this together @mattberther and sorry for the delay in reviewing!

Rather than modifying the existing `deploy.sh`script, can you make your changes in a new `travis-deploy.sh` script?

The reasons why I suggest this are:

- The `deploy.sh` comes from the upstream Slate project and when we upgrade to Slate 2.0, it will make the upgrade easier (the less we change the Slate code, the better)
- Keeping the existing `deploy.sh` will allow us to manually deploy (like we're doing today) if needed.

mattberther:
Thanks for the feedback, @kpshek. Please review, and if you're comfortable, go ahead and merge.

PR #109, opened 2017-10-30
Formally document hook definitions

### LINKED ISSUE #39: Formally document hook definitions
In discussion with @brynrhodes and @isaacvetter regarding aligning Clinical Reasoning and CDS Hooks, one of the topics of discussion was around how hooks are defined today.

For instance, the [medication-prescribe](http://cds-hooks.org/#medication-prescribe) hook is very loosely defined today with just a brief mention of the `context` data that is sent on this type of hook request.

We should do the following:

1. Strengthen our documentation around our currently predefined hooks. The key data here is the value of the `context` in the CDS Service request.
2. Clearly document expectations for how future hooks should be defined.

### LINKED ISSUE #75: Context should be an Array or an Object
Some hooks require a set of resources be passed to the decision support service, so that it can evaluate a set of actions that are inter-related. For instance, order-review - many of the uses of this hook will relate to a set of orders e.g. the set of diagnostic requests proposed. The same applies to medications particularly, where clinicians - rightly - think of a single prescription that contains multiple medications, but FHIR (also rightly) treats this as a set of resources.  

The context should be either an object or an array of objects, where each elemetn in the array is an object that has the same content as the object would have. The hook definition should define the cardinality of the context.

isaacvetter:
Fixes #39


brettmarquard:
I completely agree with this comment:

* each hook should have a complete example of the context object.

isaacvetter:
@kpshek - can you figure out why Travis won't build for this PR? I thought that it was because I had outstanding reviews from you and Bryn. But even after getting rid of those, Travis returns this:

>The command "mkdocs build --verbose --clean --strict" exited with 0.
>Skipping a deployment with the pages provider because this branch is not permitted
>Done. Your build exited with 0.

It's something to do with the travis [provider configuration](https://github.com/cds-hooks/docs/blob/iss39-formally-document-hook-definitions/.travis.yml).

kpshek:
@isaacvetter - Travis CI is building your branch just fine. Exiting with a status of 0 means that the `mkdocs build` command was successful. The "skipping deployment" message is expected because branches aren't deployed to the `gh-pages` branch. As long as you see a ✅  next to your commit hash, that means the Travis CI build succeeded.

Our review apps (that "View deployment" button) are built through Heroku, not Travis CI.

To recap:
- Travis CI builds the project via `mkdocs build`
- Travis CI will deploy the build to the `gh-pages` branch (but only if the commit is on master
- Heroku builds pull request commits and deploys them to unique review apps

kpshek:
@isaacvetter - Rather than having a `cardinality` column, I think labeling it `required` with a boolean value would be more appropriate. Cardinality has implications that don't apply here. For instance, `medication-prescribe` has a `medications-in-progress` field with a cardinality of 1..*. This implies that you could have a JSON structure like:

```javascript
{
   "patient": "123",
   "medications-in-progress": { // a FHIR medication },
   "medications-in-progress": { // another FHIR medication }
}
```

However, what we're really trying to communicate is that `medications-in-progress` is **required** and it is an `array` of FHIR medication objects. Here is a concrete proposal that tweaks what you have today:

Key | Required | JSON Type | Data |
-- | -- | -- | -- |
patient | Yes | String | FHIR Patient.id
encounter | No | String | FHIR Encounter.id
medications-in-progress | Yes | Array | DSTU2 - MedicationOrder<br />STU3 - MedicationRequest

isaacvetter:
Hey @kpshek - your example of cardinality is contrary to how the FHIR spec uses the term, but I do think that splitting the two concepts is to understand. 

By introducing "JSON data type", we now have boolean, number, null? Fundamentally, the relevant data types are:
* FHIR resource
* FHIR identifier
* string
* array 
* object
* then less importantly: number

By specifying json, we lose the primary data types.

brynrhodes:
I agree with the proposed change to add the JSON type. Yes, that means for resources it will be something like Array, with the actual resources identified by the Description, but at least with the JSON type we have the beginnings of something computable.

kpshek:
In an offline discussion with @mattberther, he is going to revert 951197b as we will make that change on #153 😄

isaacvetter:
Hey @kpshek - the example you use to differentiate pre-fetch and context is a poor example:

> +When the context data relates to a FHIR data type, it is important not to conflate context and prefetch. For instance, imagine a hook for opening a patient's chart. This hook should include the FHIR identifier of the patient whose chart is being opened, not the full patient FHIR resource. In this case, the FHIR identifier of the patient is appropriate as CDS Services may not be interested in details about the patient resource but instead other data related to this patient. Or, a CDS Service may only need the full patient resource in certain scenarios. Therefore, including the full patient resource in context would be unnecessary. For CDS Services that want the full patient resource, they can request it to be prefetched or fetch it as need from the FHIR server.

A better example would be something like - a pgx cds service will always ask for genomic marker Observations in pre-fetch, but a zika service will never use this information -- even when the same hook is used by both services.

Isaac

isaacvetter:
I really like the new prefetch context reference syntax! I think that it's important to call out that:
>Only the first level fields in context may be considered for tokens.
Means that a hook whose context includes a full FHIR resource disallows a prefetch definition from referring to a field within that resource.

As an example, if a hook's context includes a full Patient resource, it's not possible for the prefetch definition to reference the resource identifier. This can practically be worked around by always using both the full FHIR resource and a FHIR identifier as two separate elements in the context.

isaacvetter:
Pertaining to context hook definition, while it will be useful to additionally enable arbitrary key/value pairs, the primary use-case for context is and should be FHIR identifiers and FHIR resources. 

The reason that the initial PR included both a FHIR resource name and FHIR resource name is that resource name is version-specific .... This isn't addressed in the current PR. 

To recognize this, it would make sense for the documentation's example context to include a FHIR resource and/or a FHIR id.

isaacvetter:
Fyi - I opened #155 related to context on order-review

kpshek:
@isaacvetter - Thanks for the great suggestions! I've addressed each of your three suggestions as additional commits on this PR. Can you take a look now?

kpshek:
These changes also are intended to address #151

isaacvetter:
Approved!

Issue #59, opened 2017-07-19
Passing parameters between CDS Hook and SMART on FHIR app launch


yashaskram:
Passing parameters between CDS Hook and SoF app launch (when hook provider and app provider are same and would like to use the state/context

Couple of use cases 
1)	Taking users to a particular point in the SoF app based on the CDS Hook interaction. Sharing app context
2)	Pre populating information based on the hook interaction. Sharing hook id (Example Imaging appropriate use case – populating or linking back to the CDS receipt information generated in the hook interaction)

Couple of straw men (Thx to @kpshek  and @isaacvetter ) to start with. 

A link in CDS card has params data 

```json
{
  "summary": "Example Card",
  "indicator": "info",
  "detail": "This is an example card.",
  "source": {
    "label": "Static CDS Service Example",
    "url": "https://example.com"
  },
  "links": [
    {
      "label": "Google",
      "url": "https://google.com",
      "type": "smart",
      "params": "foo=a&bar=c"
    }
  ]
}
```

**SMART launch URL param**
EHR appends the params value to the SMART app launch URL. This is a change we’d want to put into SMART (to call out launch params by the SMART app) and take advantage of here in CDS Hooks.
http://<app launch url>&**foo=a&bar=c**&iss=xx&launch=xx

**OAuth claim alongside access token during SMART launch**
Similar to above, but the EHR passes the custom value in the CDS card SMART link alongside the access token during a SMART launch. In this case the value doesn’t have to be URL encoded

```json
{  
  "need_patient_banner":true,
  "patient":"123",
  "encounter":"456",
  "token_type":"Bearer",
  "access_token":"....",
  "id_token":"...",
  "scope":"patient\/Patient.read patient\/Observation.read launch online_access openid profile",
  "expires_in":570,
  "params":"foo=a&bar=c"
}
```

Other ideas / thoughts?

isaacvetter:
Hey @yashaskram, 

I think that the CDS Hooks community has an ideal that a cds service can/will recommend SMART apps that are external to it. As we get closer to production implementations, it'll be interesting to see if that's actually a common use-case. I think that it makes sense to enable tighter integration when the service and app are the same system -- as long as we continue to enable the case where they're not.

Because the SMART app launch url is typically pre-registered in an EHR's SMART app directory, the launch url itself can't be dynamic. 

[...]

PR #112, opened 2017-11-10
Add ability to pass CDS Service supplied parameter in a SMART app launch

yashaskram:
@kpshek , @isaacvetter , @brynrhodes - I have made changes based on the discussion in https://github.com/cds-hooks/docs/issues/59. Please let me your thoughts


docs/specification.md
```diff
@@ -249,8 +251,8 @@ Each **Action** is described by the following attributes.
 
 Field | Description
 ----- | -----------
-`type` |  *string*. The type of action being performed. Allowed values are: `create`, `update`, `delete`.
```
kpshek: Nit: Your changes should modify this line

kpshek:
While at FHIR DevDays last week, I talked with a couple guys from @StayWell who asked me for the same functionality described in #59 (and what this PR is attempting to address). I just emailed them to ask them to provide their feedback on the various approaches we've been discussing for their input for an additional perspective.

kpshek:
I think a new attribute on a `link` (specifically for links of type `smart` only) is the most straightforward approach to solve this particular need. It is more explicit as to the use case we are attempting to solve and works when a card has multiple SMART app links.

It is not clear to me where else an opaque string on the CDS card would be used (outside of being passed to the SMART app on launch).

yashaskram:
@kpshek Appreciate your input. I am ok to have something at the link level for this use case

>It isn't clear to me why we are adding this card-level attribute for a value that is only intended to be passed to a SMART app via a launch parameter. Can you elaborate on why the **proposal** to add a new attribute on the link wasn't ideal?

>I think a new attribute on a link (specifically for links of type smart only) is the most straightforward approach to solve this particular need 

which proposal are you referring? Can you point to this? If not, can you propose what you have is mind in issue [#59] 

For example : are you thinking of leveraging query params in the url like 
```json
{
      "cardid": "345"
      "summary": "Card 1 - Clinical Pathway 1",
      "indicator": "info",
      "detail": "This is an example app card.",
      "links": [
        {
          "label": "SMART App",
          "url": "https://smart.example.com/launch?foo=a",
          "type": "smart"
        }
      ]
    }
```

vs having an additional attribute like below. If so were you thinking extension vs dedicated attribute name

```json
{
      "cardid": "345"
      "summary": "Card 1 - Clinical Pathway 1",
      "indicator": "info",
      "detail": "This is an example app card.",
      "links": [
        {
          "label": "SMART App",
          "url": "https://smart.example.com/launch",
          "type": "smart",
          "foo": "a"
        }
      ]
    }
```

docs/specification.md
yashaskram: thx @kpshek.  I will wait to arrive at a consensus on this https://github.com/cds-hooks/docs/pull/112#issuecomment-347258271 and redo the PR

yashaskram:
@kpshek - thx for reviewing my changes, as for your comment https://github.com/cds-hooks/docs/pull/112#pullrequestreview-78423057, I will wait to arrive at a consensus on this https://github.com/cds-hooks/docs/pull/112#issuecomment-347258271 and redo the PR

kpshek:
@yashaskram - I was referring to something similar to the example noted under "A link in CDS card has params data" in #59. Here is a concrete proposal what I'm in favor of implementing here:

```json
{
  "summary": "Example card",
  "indicator": "info",
  "source": {
    "label": "Example CDS Service"
  },
  "links": [
    {
      "label": "SMART App",
      "url": "https://smart.example.com/launch",
      "type": "smart",
      "cdsHooksContext": "Some opaque (to the EHR) string that will come back alongside your OAuth 2 access token on a SMART app launch"
    },
    {
      "label": "Another SMART App",
      "url": "https://smart2.example.com/launch",
      "type": "smart",
      "cdsHooksContext": "Another context value for this SMART app"
    },
    {
      "label": "An absolute link",
      "url": "http://cds-hooks.org",
      "type": "absolute"
    }
  ]
}
```

The `cdsHooksContext` field should only ever be valued if the link `type` is `smart`. `cdsHooksContext` is not valid for `absolute` links.

The `cdsHooksContext` field and value will be sent to the SMART app alongside the other launch context when the SMART app is launched. See "Launch context arrives with your access_token" in the [SMART documentation](http://docs.smarthealthit.org/authorization/scopes-and-launch-context/). For instance, here is what would come back for the first SMART app in my example:

```javascript
{
  "access_token": "secret-xyz",
  "patient": "123",
  "cdsHooksContext": "Some opaque (to the EHR) string that will come back alongside your OAuth 2 access token on a SMART app launch"
  // etc...
}

**Comment by yashaskram** [comment]
@kpshek - I like your proposal. looks like we are back to the start point 
@isaacvetter , @brynrhodes - thoughts? based on your input I can make the changes to my PR

**Comment by isaacvetter** [comment]
Hey @yashaskram , guys,

Yeah - the above sample message from @kpshek makes sense - the cardInstance that I was pushing towards had the disadvantage of not accounted for multiple smart links in a single card. Not allowing this parameter to be specified in an absolute link makes sense because this param should be passed as a SMART launch parameter (alongside the access token). A similar parameter should simply be put into the url (perhaps in the querystring) of an absolute link.

The only part of the above proposal that I object to, is the name: `cdsHooksContext`. Ultimately, it's not descriptive of the purpose of the element. How about: `appSession` or serviceAppId` or ... something that actually describes the intent and purpose of the field.

Isaac

**Comment by yashaskram** [comment]
@isaacvetter & team - how about the name - appcontext?

**Comment by kpshek** [comment]
I struggled coming up with a descriptive name for this field as well. I originally named it `appContext` but changed it when I saw it alongside the other context variables in a SMART app launch. With that being said, I'm fine if we name this `appContext` (note the camelCase in keeping with our existing field naming conventions).

File: docs/specification/1.0.md [hunk 2]
```diff
@@ -230,8 +231,8 @@ Each **Action** is described by the following attributes.
 
 Field | Description
 ----- | -----------
-`type` |  *string*. The type of action being performed. Allowed values are: `create`, `update`, `delete`. 
-`description` | *string*. human-readable description of the suggested action. May be presented to the end-user. 
+`type` |  *string*. The type of action being performed. Allowed values are: `create`, `update`, `delete`.
```
kpshek: You shouldn't have any changes on these lines

docs/specification/1.0.md
```diff
@@ -204,8 +205,8 @@ Each **Card** is described by the following attributes.
 Field | Description
 ----- | -----------
 `summary` | *string*. one-sentence, <140-character summary message for display to the user inside of this card.
-`detail` | *string*.  optional detailed information to display, represented in [(GitHub Flavored) Markdown](https://github.github.com/gfm/). (For non-urgent cards, the EHR may hide these details until the user clicks a link like "view more details...".) 
-`indicator` | *string*.  urgency/importance of what this card conveys. Allowed values, in order of increasing urgency, are: `info`, `warning`, `hard-stop`. The EHR can use this field to help make UI display decisions such as sort order or coloring. The value `hard-stop` indicates that the workflow should not be allowed to proceed. 
+`detail` | *string*.  optional detailed information to display, represented in [(GitHub Flavored) Markdown](https://github.github.com/gfm/). (For non-urgent cards, the EHR may hide these details until the user clicks a link like "view more details...".)
```
kpshek: You shouldn't have any changes on these lines.

docs/specification/1.0.md
```diff
@@ -241,6 +242,8 @@ Field | Description
 <nobr>`label`</nobr>| *string*. human-readable label to display for this link (e.g. the EHR might render this as the underlined text of a clickable link).
 `url` | *URL*. URL to load (via `GET`, in a browser context) when a user clicks on this link. Note that this may be a "deep link" with context embedded in path segments, query parameters, or a hash.
 `type` | *string*. The type of the given URL. There are two possible values for this field. A type of `absolute` indicates that the URL is absolute and should be treated as-is. A type of `smart` indicates that the URL is a SMART app launch URL and the EHR should ensure the SMART app launch URL is populated with the appropriate SMART launch parameters.
+<nobr>`appContext`</nobr> |*string*.  an optional field to share addtional app context from CDS Service to EHR. The `appContext` field should only be valued if the link type is `smart` and is not valid for `absolute` links. The `appContext` field and value will be sent to the SMART app alongside the other launch context when the SMART app is launched.
```
kpshek: I think it would be more clear to wordsmith the first sentence to (also fixed a typo):

"An optional field that allows the CDS Service to pass context regarding the launch of this SMART app from the CDS card to the SMART app."

Also, you don't need the `<nobr>` tags around `appContext`

docs/specification/1.0.md
```diff
@@ -296,7 +299,7 @@ Regardless of how the EHR satisfies the prefetched templates (if at all), it is
 
 The resulting response, which must be rendered in a single page — no "next
 page" links allowed — is passed along to the CDS Service using the
-`prefetch` parameter (see below for a complete example).
```
kpshek: You shouldn't have any changes on this line.

kpshek:
@travistcummings - can the HSPC Sandbox support passing arbitrary launch context to the SMART app per these proposed changes?

Tagging @zplata for the eventual [cds-hooks/sandbox](https://github.com/cds-hooks/sandbox) change that this will also require.

docs/specification/1.0.md
yashaskram: there are trailing whitespace in couple of lines in this file. Not sure why but the courtesy of atom whitespace package automagically removes them when saving the file based on package setting and I let it. But I have changed it back now to move forward with this PR.

docs/specification/1.0.md
yashaskram: I like your version. I have updated text based on your input. 

as for <nobr> , there are many places in this file which leverages this tag to prevent wrap .  I have removed this based on your comment but something to thing about

docs/specification/1.0.md
kpshek: Thanks @yashaskram. fyi, if you use `git add --interactive` you can add just the changes you want in a file rather than the entire file. Also, I know some Git GUI programs support adding individual hunks as well.

travistcummings:
@kpshek The HSPC Sandbox launch will support additional parameters in the SOF launch url: https://example.com/launch.html?p1=abc&p2=xyz

When the CDS Hooks sandbox registers a launch context, it can also provide a params map object that will be passed to the client along side their access token:

launch context:
{ 
...
"parameters":{ 
"patient":"BILIBABY",
"appContext": {
"p1": "abc",
"p2": "xyz"
},
"need_patient_banner":false
}
}

SOF launch response:
{
"access_token": "eyJhbGc...",
"token_type": "Bearer",
"expires_in": 86399,
"scope": "patient/Observation.read launch patient/Patient.read patient/Observation.write",
"patient": "BILIBABY",
"need_patient_banner": false,
"appContext": {
"p1": "abc",
"p2": "xyz"
}
}

kpshek:
That's great, thanks @travistcummings!

mtcollie:
Hi @kpshek , apologies for the delay, but this parameter would be just what we need to provide contextual links to particular content or a program.  For example, if a physician prescribed a sleep aid medication, we could prompt with a notification that the patient was eligible to be enrolled in the sleep health program and the link to could take them directly to the enrollment section.

PR #115, opened 2017-11-19
Add support for versioning the specification

### LINKED ISSUE #100: Version CDS Hooks specification
We need to version the CDS Hooks specification and deploy each version to its own distinct URL

kpshek:
This PR adds support for versioning the CDS Hooks specification. Today, the specification lives at `/specification`. With this change, it is now at `/specification/1.0`.

Future versioning can live at similar URLs. Eg:
- `/specification/1.1`
- `/specification/2.0`

This fixes #100

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -1,5 +1,8 @@
 # CDS Services
 
+!!! info "1.0 Draft"
+    This is the draft of the 1.0 release of the CDS Hooks specification. We are currently working towards a 1.0 release and would love your feedback and proposed changes. Look at our <a href="http://github.com/cds-hooks/docs/issues">current issue list</a> and get involved!
+
 ## Discovery
 
 ```shell
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -12,7 +12,8 @@ extra:
 
 pages:
 - Overview: 'index.md'
-- Specification: 'specification.md'
+- Specification:
+    - '1.0': 'specification/1.0.md'
 - Hooks: 'hooks.md'
 - Examples: 'examples.md'
 - Community: 'community.md'


isaacvetter:
@kpshek , in 1.0.md, I think that you should add the uuid back in suggestions. Note that the AUC use-case requires tracking an externally recommended suggestion for auditing purposes via a unique identifier.

kpshek:
> Thanks, this makes sense. Did you consider skipping the dropdown for now?
> Just - Specification: 'specification/1.0.md' so link takes you directly to 1.0 page?

@brettmarquard - I did but this isn't possible in mkdocs today. They have a long standing issue logged for this ([mkdocs/mkdocs #73](https://github.com/mkdocs/mkdocs/issues/73)).

After some thought, I like the proposed approach here better. This forces everyone to always link to a versioned specification. A common thing I see with the FHIR homepage URL structure is that people often link to the versionless page, not realizing that in the future that will be updated.

vadi2:
Definitely agree, this is a good move to make it explicit.

jmandel:
I'm happy with the explicit treatment. A few notes:

*  `specification` page it's not the only part of the site that might change with different versions. The examples certainly would, and the hooks catalog might. Especially if we have backwards incompatible changes in the future.

* It might be good to be explicit about the current draft status too, like `specification-1.0-draft` (similar to SNAPSHOT in the maven world).

kpshek:
Thanks for the feedback, @jmandel!

> specification page it's not the only part of the site that might change with different versions. The examples certainly would, and the hooks catalog might. Especially if we have backwards incompatible changes in the future.

Agreed. My thinking is that we address those changes/pages when we have > 1 version. Given that we won't be "releasing" that content, I didn't think it was a pressing need to address at this point.

> It might be good to be explicit about the current draft status too, like specification-1.0-draft (similar to SNAPSHOT in the maven world).

I considered this now but thought that I would just leave it as-is with the note at the top of the page. Once we release 1.0 we'll do this for the next version (eg, `1.1-draft.md`)

PR #116, opened 2017-11-19
Remove decisions for the 1.0 release

### LINKED ISSUE #99: Remove Decisions for the 1.0 release
Based upon a lack of broad implementer feedback and due to the number of outstanding issues around Decisions, a consensus was reached last week in San Diego that Decisions will not be included in the 1.0 release.

Decisions are an important feature of CDS Hooks that we will want maintain. To include this in CDS Hooks, it is imperative that we receive broad implementer feedback on this feature and thus far, we have not seen this. Additionally, we need to address many, if not all of the outstanding issues that have been logged against Decisions.

Our plan is to circle back on Decisions after 1.0 and include it in a subsequent release of CDS Hooks.

kpshek:
Remove decisions and the `redirect` parameter in preparation for a 1.0 release.

Fixes #99.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -36,22 +36,8 @@ materials (*app link card*). A user sees these cards — one or more of each typ
 
 * *suggestion card*: provides a specific suggestion for which the EHR renders a button that the user can click to accept. Clicking automatically populates the suggested change into the EHR's UI.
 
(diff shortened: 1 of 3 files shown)


kpshek:
@isaacvetter - Why did you remove that link the example in 9678739? That doesn't have anything to do with the decisions feature. Additionally, that additional link was to show that you can have multiple `absolute` links associated with a card.

PR #118, opened 2017-11-19
Add Open API Specification (Swagger)

### LINKED ISSUE #117: Add Swagger API
Currently, the Swagger API definition is defined in a separate repository ([cds-hooks/api](https://github.com/cds-hooks/api)). We should merge this into this repository so that:

- We can easily modify / keep the Swagger definition up-to-date anytime the specification changes
- We can version the Swagger API definition along with the specification

kpshek:
Merge the Open API Specification (Swagger) from the [cds-hooks/api](https://github.com/cds-hooks/api) repository into this repository.

Note that I've made this change on top of the specification versioning change to make it easier once that change gets merged in.

Fixes #117.

### DIFF
--- a/docs/specification/1.0-api.yaml
+++ b/docs/specification/1.0-api.yaml
@@ -1,0 +1,219 @@
+swagger: '2.0'
+
+################################################################################
+#                              API Information                                 #
+################################################################################
+info:
+  version: '1.0'
+  title: CDS Hooks REST API
+host: cds-service-example.herokuapp.com
+schemes:
+  - https
+consumes:
+  - application/json
+produces:
+  - application/json
+
+################################################################################
+#                                           Paths                              #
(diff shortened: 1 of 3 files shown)


kpshek:
> I like the yaml, but it's not clear to me whether/how the documentation would/could be derived from that? By having the yaml, will we need to synchronize that definition whenever things change, or is there a way to drive the specification docs from the yaml?

@brynrhodes - We're always going to have the Markdown + YAML. The YAML allows you to have some documentation as well but mkdocs/Markdown allows a lot more flexibility in the documentation that just isn't possible with the YAML.

The big benefit of moving the `api` repository that held this YAML here is that now when we make a change to the specification, we can be sure that both the Open API specification and the mkdocs content is updated together.

PR #126, opened 2017-12-20
Service Specification

bakerdb:
I've tightened up the table showing the fields in a service description such that each field has an associated "priority" of REQUIRED, RECOMMENDED, or OPTIONAL (not used in this example).  This change illustrates my recommendation that the 1.0 spec needs to be specified more explicitly.  In addition, it carries out my recommendation to make the Prefetch field RECOMMENDED rather than required, so that each provider can decide whether to prefetch or allow the CDS service to fetch its own data as needed.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -58,18 +58,18 @@ The discovery endpoint is always available at `{baseUrl}/cds-services`. For exam
 The response to the discovery endpoint is an object containing a list of CDS Services.
 
 Field | Description
------ | -----------
+----- | ---------
 `services` | *array*. An array of **CDS Services**
 
 Each CDS Service is described by the following attributes.
 
-Field | Description
------ | -----------
-`hook`| *string* or *url*. The hook this service should be invoked on. See [Hook Catalog](#hook-catalog)
-`title`| *string*.  The human-friendly name of this service
-<nobr>`description`</nobr>| *string*. The description of this service
-`id` | *string*. The {id} portion of the URL to this service which is available at<br />`{baseUrl}/cds-services/{id}`
-`prefetch` | *object*. An object containing key/value pairs of FHIR queries to data that this service would like the EHR prefetch and provide on<br />each service call. The key is a *string* that describes the type<br />of data being requested and the value is a *string* representing<br />the FHIR query.<br />(todo: link to prefetching documentation)
+Field | Priority | Description
+----- | ----- | ---------
+`hook`| REQUIRED | *string* or *url*. The hook this service should be invoked on. See [Hook Catalog](#hook-catalog)
+`title`| RECOMMENDED | *string*.  The human-friendly name of this service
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -58,18 +58,18 @@ The discovery endpoint is always available at `{baseUrl}/cds-services`. For exam
...
 Each CDS Service is described by the following attributes.
 
-Field | Description
------ | -----------
-`hook`| *string* or *url*. The hook this service should be invoked on. See [Hook Catalog](#hook-catalog)
-`title`| *string*.  The human-friendly name of this service
-<nobr>`description`</nobr>| *string*. The description of this service
-`id` | *string*. The {id} portion of the URL to this service which is available at<br />`{baseUrl}/cds-services/{id}`
-`prefetch` | *object*. An object containing key/value pairs of FHIR queries to data that this service would like the EHR prefetch and provide on<br />each service call. The key is a *string* that describes the type<br />of data being requested and the value is a *string* representing<br />the FHIR query.<br />(todo: link to prefetching documentation)
+Field | Priority | Description
```
kpshek: Can you rename `Priority` to `Required`? This will match what @isaacvetter has been doing in PR #109 to indicate the same thing for the fields in each hook definition.

docs/specification/1.0.md
```diff
@@ -58,18 +58,18 @@ The discovery endpoint is always available at `{baseUrl}/cds-services`. For exam
...
+----- | ----- | ---------
+`hook`| REQUIRED | *string* or *url*. The hook this service should be invoked on. See [Hook Catalog](#hook-catalog)
```
kpshek: Rather than REQUIRED, RECOMMENDED, and OPTIONAL, I think we should communicate just whether a field is required or not (`yes` or `no`), similar to PR #109.

The following fields should be required (indicated by a `yes` value in the column cell): `hook`, `title`, `description`, `id`.

`prefetch` is an optional field (indicated by a `no` value in the column cell). I do not think the specification should state that this field is recommended. There are certainly CDS Services that do not need or want any prefetched data.

bakerdb:
I can see the wisdom of making "prefetch" OPTIONAL rather than RECOMMENDED.  Agreed that not all CDS services will need prefetched data.
I do not agree with the recommendation to limit priorities to REQUIRED or not (yes/no).  This specification should use the conformance language specified in RFC 2119 (https://www.ietf.org/rfc/rfc2119.txt), just as the FHIR specification and SMART on FHIR specifications (and most other specifications) do.  The nuanced difference between "not required" and OPTIONAL or RECOMMENDED is important to capture.

kpshek:
@bakerdb - Fair enough. I'm fine with specifying either `REQUIRED` or `OPTIONAL` for the fields here. I also agree that if we have something that is `RECOMMENDED` (though we don't here), we can call that out.

kpshek:
In an offline conversation with @bakerdb, she is going to update `prefetch` to `OPTIONAL` and then we should be good on this PR.

PR #127, opened 2017-12-22
Update Discovery documentation

bakerdb:
Changes to clarify discovery of CDS services.  
Also to serve as the set-up for a new section on Service Registration -- to come.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -13,7 +13,9 @@ The CDS Hooks specification is available as an Open API Specification using Swag
 
 ## Discovery
 
-In order to allow EHRs to discover its services, a CDS service needs to provide a stable endpoint.  A CDS service MAY expose its services at 
+Developers of CDS Services MUST provide a stable endpoint for allowing EHRs to discover available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
+
+A CDS service MAY expose its services at 
 
 ```shell
 {baseURL}/cds_services  
@@ -52,8 +54,6 @@ curl "https://example.com/cds-services"
   ]
 }
 ```
-
-Developers of CDS Services MUST provide a stable endpoint for allowing EHRs to discover available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
 
 ### HTTP Request
 
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -13,12 +13,12 @@ The CDS Hooks specification is available as an Open API Specification using Swag
 
 ## Discovery
 
-Developers of CDS Services MUST provide a stable endpoint for allowing EHRs to discover available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -45,7 +53,7 @@ curl "https://example.com/cds-services"
 }
 ```
 
-Developers of CDS Services must provide a well-known endpoint allowing the EHR to discover all available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
+Developers of CDS Services MUST provide a stable endpoint for allowing EHRs to discover available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
```
kpshek: This sentence just needs to be pulled up to the top of the `Discovery` section. When we were using another documentation templating engine (Slate), our code examples had to proceed the text that it was in reference to. With the switch to MkDocs, our examples are now in-line which causes this to look odd.

Moving this sentence to the top of the section will make the new sentence you added above redundant I believe.

Out of curiosity, why do you like "stable" more than "well-known"? The word "well-known" to me better describes why we are prescriptive of the `/cds-services` endpoint. The word "stable" to me only means that an endpoint must be accessible/available rather than at some predictable location.

docs/specification/1.0.md
```diff
@@ -13,6 +13,14 @@ The CDS Hooks specification is available as an Open API Specification using Swag
 
 ## Discovery
 
+In order to allow EHRs to discover its services, a CDS service needs to provide a stable endpoint.  A CDS service MAY expose its services at
```
kpshek: Per my below comment, I think this sentence is redundant once we move the original sentence up here.

What value do you see the additional example adding given that we have this same statement below as well? Additionally, it is `cds-services` (dash), not `cds_services` (underscore).

docs/specification/1.0.md
```diff
@@ -13,6 +13,16 @@ The CDS Hooks specification is available as an Open API Specification using Swag
 
 ## Discovery
 
+Developers of CDS Services MUST provide a stable endpoint for allowing EHRs to discover available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
+
+A CDS service MAY expose its services at 
+
+```shell
+{baseURL}/cds_services
```
kpshek: This should be a dash, not underscore: `{baseURL}/cds-services`

docs/specification/1.0.md
```diff
@@ -13,6 +13,16 @@ The CDS Hooks specification is available as an Open API Specification using Swag
 
 ## Discovery
 
+Developers of CDS Services MUST provide a stable endpoint for allowing EHRs to discover available CDS Services, including information such as the purpose of the CDS Service, when it should be invoked, and any data that is requested to be prefetched.
+
+A CDS service MAY expose its services at
```
kpshek: Tagging onto what Bryn pointed out, perhaps we reword this simply to "A CDS Service provider MUST expose its Discovery endpoint at"

kpshek:
Thanks @bakerdb for the updates! Once @brynrhodes re-reviews we are good to go here!

PR #132, opened 2017-12-22
Prefetch Clarifications

bakerdb:
Clarification of issues discussed w.r.t. prefetch, it's need to be pre-negotiated, and the architectural dependence of its implementation.  Responds to risk R3.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -139,12 +139,11 @@ Note: the `hookInstance` is globally unique and should contain enough entropy to
 
 #### prefetch
 
-As a performance tweak, the EHR may pass along data according
+As a performance tweak, the EHR MAY pass along data according
 to the service's [[Prefetch-Template]]. This helps provide the service with all
 the data it needs to efficiently compute a set of recommendations. Each key matches a key described in the CDS Service Discovery document; and each value is a FHIR Bundle.entry indicating a response status and returned resource.
 
-Note that in the absence of `prefetch`, an external service can always execute
-FHIR REST API calls against the EHR server to obtain additional data ad-hoc.)
+Note that in the absence of `prefetch`, a CDS service MAY present the access token passed by the EHR in the `fhirAuthorization` parameter to retrieve additional FHIR resources from the EHR FHIR server.   
 
 <aside class="notice">
 You can see the <a href="http://editor.swagger.io/?url=https://raw.githubusercontent.com/cds-hooks/api/master/cds-hooks.yaml">complete data model in Swagger</a>.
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -139,12 +139,11 @@ Note: the `hookInstance` is globally unique and should contain enough entropy to
...
 #### prefetch
 
-As a performance tweak, the EHR may pass along data according
+As a performance tweak, the EHR MAY pass along data according
 to the service's [[Prefetch-Template]]. This helps provide the service with all
 the data it needs to efficiently compute a set of recommendations. Each key matches a key described in the CDS Service Discovery document; and each value is a FHIR Bundle.entry indicating a response status and returned resource.
 
-Note that in the absence of `prefetch`, an external service can always execute
-FHIR REST API calls against the EHR server to obtain additional data ad-hoc.)
+Note that in the absence of `prefetch`, a CDS service MAY present the access token passed by the EHR in the `fhirAuthorization` parameter to retrieve additional FHIR resources from the EHR FHIR server.
```
brynrhodes: Use of the "fhirAuthorization" token in this context is incidental so it seems like a distraction to mention it here. Suggest retaining the original working, or something like:

Note that if `prefetch` data is not provided as part of the request, the service can always execute FHIR REST API calls against the provided FHIR URL to obtain additional data to process the request.

docs/specification/1.0.md
```diff
@@ -273,35 +272,37 @@ If a suggestion has no `uuid`, the EHR does not send a notification.
...
+the data minimally necessary to complete the requested decision-making process, where the following variables are defined:
 
 |variable|meaning|
 ---------|--------
 |`{{Patient.id}}`|The id of the patient in context for this activity (e.g. `123`)|
 |`{{User.id}}`|The type and id of the user for this session (e.g. `Practitioner/123`)|
 
-An EHR *may* choose to honor some or all of the desired prefetch templates from an appropriate source. For example:
+The CDS Hooks specification necessarily does not constrain how “prefetch” is implemented.  Whether and how prefetch is implemented is left up to the implementer, as architecturally dependent, and involves out-of-band negotiation between the health system, the EHR vendor, and the CDS service.  The prefetch MUST NOT exceed the user-based authorizations for the context from which the CDS service is called, and it MUST NOT exceed the data minimally necessary to complete the requested decision-making.  The EHR MAY  NOT prefetch any FHIR resources beyond those for which the current user is authorized.  
+
```
brynrhodes: Agree with the constraints around only providing data that the user is authorized to access, but disagree with the constraint that the EHR only provide data that is "minimally necessary" to complete the requested decision-making. That's a significant burden to place on the implementation and it would only apply to PHI, which is only that information that can actually be used to identify a patient.

docs/specification/1.0.md
kpshek: I think the question here is around the opening line to this sentence which may give the impression that `fhirAuthorization` is only used if the prefetch request is not honored. Perhaps we simply reword this to:

"In addition to requesting data through `prefetch`, a CDS service..."

docs/specification/1.0.md
kpshek: @bakerdb - The concern here is with the statement:

>and it MUST NOT exceed the data minimally necessary to complete the requested decision-making.

Was the intention to address [HIPAA's minimum necessary requirement](https://www.hhs.gov/hipaa/for-professionals/privacy/guidance/minimum-necessary-requirement/index.html)?

Or, was this statement trying to state that if a CDS Service wants the patient's current medications prefetched they should be given something unrelated like labs?

Bryn's concern (which I think is valid) is a scenario like this: Imagine a CDS Service needs just the medication name for its algorithm. It requests the patient's current medications to be prefetched. The EHR satisfies the prefetch request by sending the full FHIR medication resource.

Did the EHR violate the CDS Hooks standard by sending the full medication resource rather than just a paired down minimal medication resource composed of just the name? I would say no -- I have not seen discrete/granular control over individual data fields in systems I've worked on. As an aside, doing this detail level of fine-grain control over the fields is not possible with SMART on FHIR's scopes nor is it being implemented with FHIR implementations today in production.

docs/specification/1.0.md
bakerdb: I changed the "Note that..." to Kevin's suggested wording.  
However, I don't agree that a service "can always..." -- a service can retrieve data through the FHIR API only if the EHR has provided an access token.  I have reworded accordingly.

docs/specification/1.0.md
bakerdb: Because the data are provided within the context of a single patient, any data prefetched or available through an access token passed within that context are by definition "PHI."  However, I don't think the CDS Hooks specification should be a HIPAA-compliance mechanism.  So I have softened this language.

bakerdb:
I tried to resolve the conflicts (deleted all from patch-8, retaining master), but saw no way to save these decisions.  Maybe I don't have the right access privileges?

kpshek:
> I tried to resolve the conflicts (deleted all from patch-8, retaining master), but saw no way to save these decisions. Maybe I don't have the right access privileges?

You should be able to resolve the conflicts yourself by update your branch against the latest master. However, I can just resolve the conflicts when I merge this to master too.

bakerdb:
I could delete the patch-8 options (leaving the master options) in #132, but the “Mark as Resolved” button is not active.

-Dixie


Dixie B. Baker, Ph.D.
Senior Partner
Martin, Blanck and Associates
Office (Redondo Beach, CA):  310-791-9671
Mobile:  310-279-2579

On Jan 24, 2018, at 7:26 AM, Kevin Shekleton <notifications@github.com<mailto:notifications@github.com>> wrote:


I tried to resolve the conflicts (deleted all from patch-8, retaining master), but saw no way to save these decisions. Maybe I don't have the right access privileges?

You should be able to resolve the conflicts yourself by update your branch against the latest master. However, I can just resolve the conflicts when I merge this to master too.

kpshek:
With all of the conflicts between these changes and #136 and my errors in trying to resolve the merge conflicts, I think it will be easier to simply review these changes on a clean PR: #145.

PR #135, opened 2018-01-04
Document CDS Services should abide by accessing only the minimum necessary data neeeded

bakerdb:
Added clarification that the data elements a CDS service requests to be prefetched needs to be the "minimum necessary."  
Added priorities to CDS description table.  Responsive to risk #R16 from CDS service perspective.

Note that the table includes the statement "(todo: link to prefetching documentation)."  If this documentation does not exist, suggest deleting this statement.  Else add the link.


kpshek:
In an offline conversation with @bakerdb, we are going to wait on this PR until #126 is merged to master since some of these changes coincide.

PR #137, opened 2018-01-05
Improve FHIR Resource Access documentation

bakerdb:
Incorporates all of the concepts we have discussed w.r.t. enabling an EHR to obtain a bearer token on behalf of a CDS service, and passing it on to that service in the service request.  I moved all of the pre-existing discussion re SMART into to introductory section so that the specification sections could be purely specification.  

I discovered that RFC7521 defines the "subject" parameter as "When using assertions as an authorization grant, the Subject identifies an authorized accessor for which the access token is being requested (typically, the resource owner or an authorized delegate)."  This seems to fit our purposes nicely, so I have used that as the identifier of the CDS Service that will be using the token.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -445,46 +445,65 @@ At this time, CDS Hooks does not prescribe how the EHR shares its public key or 
 
 ### FHIR Resource Access
 
-The CDS Service is able to use the FHIR server of the EHR to obtain any additional data it needs in order to perform its decision support. This is similar to SMART on FHIR where the SMART app can obtain additional data via the provided FHIR server.
+The CDS Service is able to use the EHR's FHIR server to obtain any FHIR resources it requires beyond those provided by the EHR as prefetched data. This is similar to the approach used by SMART on FHIR wherein the SMART app requests and ultimately obtains an access token from the EHR's Authorization server using the SMART launch workflow, as described in [Smart App Authorization Guide](http://docs.smarthealthit.org/authorization/).  This process utilizes the authorization code grant model as defined by the [OAuth 2.0 Authorization Framework (RFC6749)](https://tools.ietf.org/html/rfc6749).
 
-Like SMART on FHIR, CDS Hooks requires that access to the FHIR server be controlled by an Authorization server utilizing the OAuth 2 framework. Thus, the CDS Service is able to consume the given FHIR server via an access (bearer) token just like a SMART app. While CDS Hooks shares the underlying technical framework and standards as SMART on FHIR, there are very important differences between SMART and CDS Hooks.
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -441,46 +441,61 @@ The EHR’s authorization server is responsible for enforcing restrictions on th
...
 
-With CDS Hooks, the EHR provides the access token directly in the request to the CDS Service. Thus, the CDS Service does not need to request the token from the authorization server as a SMART app would. This is done purely for performance reasons as the authorization code grant model in OAuth 2 involves several HTTPS calls and redirects. In contrast to a SMART app, a CDS Service may be invoked many times during a workflow. Going through the authorization code grant model on every hook invocation would likely result in a poorly performing CDS Service due to the authorization overhead.
+The authorization server’s response MUST indicate that the access token MAY NOT be cached or otherwise persisted in data storage.  The expiry time for which the token remains value SHOULD BE set as short as possible to provide the requested decision support.
 
 ```json
 {
   "fhirAuthorization" : {
     "access_token" : "some-opaque-fhir-access-token",
-    "token_type" : "Bearer",
+    "token_type" : "bearer",
```
kpshek: Based upon [rfc 6750](https://tools.ietf.org/html/rfc6750#page-14), I believe the term is "Bearer" with a capital B.

docs/specification/1.0.md
```diff
@@ -441,46 +441,61 @@ The EHR’s authorization server is responsible for enforcing restrictions on th
...
     "expires_in" : 300,
-    "scope" : "patient/Patient.read patient/Observation.read"
+    "scope" : "patient/Patient.read patient/Observation.read",
+    "subject" : "cds-service4",
+    "aud" : "http://fhirServer.ehr.com
```
kpshek: We don't use the `aud` parameter. We discussed adding in the JWT but ultimately decided against it since every CDS Service request has a `fhirServer` field which is duplicative here.

docs/specification/1.0.md
```diff
@@ -441,46 +441,61 @@ The EHR’s authorization server is responsible for enforcing restrictions on th
...
-
-- The CDS Service being invoked
-- The current user
+Field | Priority | Description
+----- | ----- | -----------
+`access_token` | REQUIRED | *string*. This is the OAuth 2 access token that provides access to the FHIR server.
+`token_type`   | REQUIRED | *string*. Fixed value: `Bearer`
+`expires_in`   | REQUIRED | *integer*. The lifetime in seconds of the access token.
+`scope`        | REQUIRED | *string*. The scopes the access token grants the CDS Service.
+`subject` | REQUIRED | *string*.  The CDS Service as registered with the EHR.
```
kpshek: Would "The OAuth 2 client identifier of the CDS Service" be a better description of `subject`?

docs/specification/1.0.md
bakerdb: @kpshek  Agreed.

PR #139, opened 2018-01-11
New Security and Safety Content

bakerdb:
New and revised content for Security and Safety section from beginning up to "FHIR Resource Access," whose changes are addressed in PR#137.  
Also moved example to follow parameters chart.


docs/specification/1.0.md
```diff
@@ -383,35 +383,64 @@ To reduce the implementation burden on EHRs that support CDS services, CDS Hooks
...
+Once a CDS Service provider is selected, the EHR vendor/provider negotiates the terms under which service will be provided.  This negotiation includes agreement on patient data elements that will be prefetched and provided to the CDS Service, data elements that will be made available through an access token passed by the EHR, and steps the CDS Service must take to protect patient data and access tokens.  To enable the EHR authorization server to authorize CDS Service access to FHIR resources, the CDS Service is registered as a client to the EHR authorization server.  These business agreements are documented in the service agreement.  Because the CDS Service will receive patient data and will provide clinical decision support, the Service provider also will need to sign a Business Associate Agreement (BAA) that commits it to compliance with the Privacy and Security Rules of the Health Insurance Portability and Accountability Act (HIPAA).   
+
```
brynrhodes: It seems to me that this is not something the specification needs to be prescriptive about. Perhaps more general language is appropriate here:

The EHR and CDS Service will need to take whatever steps are necessary in their regulatory environments to ensure that the appropriate agreements are in place to cover usage of patient information.

docs/specification/1.0.md
```diff
@@ -383,35 +383,64 @@ To reduce the implementation burden on EHRs that support CDS services, CDS Hooks
...
+The service agreement negotiated between the EHR vendor/provider and the CDS Service provider will include obligations the EHR vendor/provider commits to the CDS Service provider. Some agreements may include the use of mutual TLS, in which both ends of the channel are authenticated.  
+
+However, mutual TLS is impractical for many organizations, and because the EHR initiates the TLS channel set-up, only the CDS Service endpoint will be authenticated.  To enable the CDS Service to authenticate the identity of the EHR, CDS Hooks uses digitally signed [JSON web tokens (JWT)](https://jwt.io/).  
+
+Each time an EHR transmits a request to a CDS Service, the request MUST include an `Authorization` header presenting the JWT ` as a “Bearer” token:
+```
+Authorization:  Bearer {{JWT}}
+```
+Note that this is for every single CDS Service call, whether that be a Discovery call, a single CDS Service invocation, or multiple exchanges relating to a single service.   Also note that mutual TLS MAY be used alongside JSON web tokens to establish trust of the EHR by the CDS Service.
```
brynrhodes: Is it really the case that Discovery calls need to be secured with a JWT?

docs/specification/1.0.md
```diff
@@ -383,35 +383,64 @@ To reduce the implementation burden on EHRs that support CDS services, CDS Hooks
...
+
+Every interaction between an EHR and a CDS Service is initiated by the EHR sending a service request to a CDS Service endpoint protected using the [Transport Layer Security protocol](TLS, RFC5246)[ https://tools.ietf.org/html/rfc5246].   Through the TLS protocol the identity of the CDS Service is authenticated, and an encrypted transmission channel is established between the EHR and the CDS Service. Both the Discovery endpoint and individual CDS Service endpoints are TLS secured.
```
brynrhodes: This paragraph implies that mutual TLS is required, but subsequent paragraphs make clear that mutual TLS is optional as it's a function of the security environment. Suggest dropping the last sentence.

docs/specification/1.0.md
kpshek: @brynrhodes - Yes, our intention with the JWT was that it is present in every request to the CDS Service, including Discovery calls.

docs/specification/1.0.md
kpshek: @bakerdb - The comment from @brynrhodes is regarding this sentence:

"Because the CDS Service will receive patient data and will provide clinical decision support, the Service provider also will need to sign a Business Associate Agreement (BAA) that commits it to compliance with the Privacy and Security Rules of the Health Insurance Portability and Accountability Act (HIPAA)."

This is a US-centric statement and something we'll want to avoid given that CDS Hooks is a global standard.

Is there a more general, global way of stating this? Or, can we just drop this sentence?

docs/specification/1.0.md
kpshek: @brynrhodes - I don't see this paragraph talking about _mutual_ TLS. It reads to me that we are just talking about TLS here.

Can you elaborate on the part that makes you think it is regarding _mutual_ TLS?

brynrhodes: Oh, you're right, disregard, I read the "both" as somehow applying to the EHR and CDS Service, but I see they are both endpoints exposed by the CDS Service.

docs/specification/1.0.md
brynrhodes: That seems like overkill to me, but okay. I am curious what the rationale is, but I don't feel strongly about this one way or the other.

kpshek: For starters, it shouldn't be any additional work for EHRs to construct the JWT. If they are already constructing it for CDS Service requests, they can do it for Discovery requests just as easily.

Second (and the primary reason), I can see some CDS Service provider not wanting any of their endpoints being called (including Discovery) unless from a trusted EHR.

docs/specification/1.0.md
bakerdb: @kpshek @brynrhodes  I agree that the sentence re HIPAA is US-specific.  Since a commitment to protect sensitive data is included earlier in this paragraph, I will delete the last sentence.

docs/specification/1.0.md
bakerdb: @kpshek @brynrhodes  Right, server-point TLS is all that is needed to conform to the assertion in the last sentence.

PR #113, opened 2017-11-15
Add initial 'Quick Start' page

brettmarquard:
this quick start is based on the Argonaut tutorial

PR #140, opened 2018-01-11
Cross-Origin Resource Sharing

bakerdb:
Replaced most of this content with simple statement regarding CORS and a reference to the OWASP guidelines.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -488,9 +488,9 @@ As the CDS Service is executing on behalf of a user, it is important that the da
 
 ### Cross-Origin Resource Sharing
 
-[Cross-origin resource sharing (CORS)](https://www.w3.org/TR/cors/) is a W3C standard mechanism that uses additional HTTP headers to enable a web browser to gain permission to access selected resources from a server in an Internet domain different from that of the site currently in use.  In short, CORS allows server APIs to control how browsers access resources on the server, including the accessible HTTP response headers. CORS is only honored by web browsers and as such, is a client-side security mechanism.  Security risks associated with CORS are well documented.
+[Cross-origin resource sharing (CORS)](https://www.w3.org/TR/cors/) is a W3C standard mechanism that uses additional HTTP headers to enable a web browser to gain permission to access resources from servers in an Internet domain outside that in which the browser is used.  CORS is a client-side security mechanism with well-documented security risks.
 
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -484,21 +484,9 @@ As the CDS Service is executing on behalf of a user, it is important that the da
...
-An insecure, extremely permissive, non-production CORS configuration that allows you to test your CDS Service against our Sandbox returns the following HTTP headers:
-
-Header | Value
------- | -----
-Access-Control-Allow-Credentials | true
-Access-Control-Allow-Headers | Authorization, Content-Type
-Access-Control-Allow-Methods | GET, POST, OPTIONS
-Access-Control-Allow-Origin | *
-Access-Control-Expose-Headers | Origin, Accept, Content-Location, Location, X-Requested-With
+Because CDS Services may return decisions involving multiple domains (e.g., link to SMART app),  browser-based EHRs will require CORS support in order to access these services.  A secure implementation guide for CORS is outside of the scope of this CDS Hooks specification. Organizations planning to implement CDS Hooks with CORS support are referred to the Cross-Origin Resource Sharing section of the [OWASP HTML5 Security Cheat Sheet]( https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet).
```
brynrhodes: The example configuration here was extremely useful, and clearly marked as non-production. Do we really need to exclude it from the documentation?

kpshek: The concern @bakerdb had was that our non-secure example was in the specification. I agree this was valuable to help developers get started, so perhaps we should just move this to our (soon to be released) quick start guide (#113)?

docs/specification/1.0.md
```diff
@@ -491,21 +491,9 @@ As the CDS Service is executing on behalf of a user, it is important that the da
...
```
kpshek: > Because CDS Services may return decisions involving multiple domains (e.g., link to SMART app),  browser-based EHRs will require CORS support in order to access these services.

CORS isn't required due to links to SMART apps from other domains. Rather, CORS must be handled by EHRs choosing to implementing CDS Hooks in a browser (CORS enables the EHR to make a CDS Service request and read the response/headers). CORS should also be implemented by CDS Services such that browser based EHRs can call them.

Perhaps we replace this first sentence with simply:  "CDS Services and browser-based EHRs will require CORS support."

docs/specification/1.0.md
bakerdb: @brynrhodes @kpshek stated my concern.  I would not advocate hiding the fact that the Sandbox requires CORS, but I don't think the CDS Hooks specification is the right place to discuss the Sandbox.  Quick Start Guide may be a better place to put this information.

docs/specification/1.0.md
bakerdb: @kpshek  In thinking further about this, it occurred to me that since CORS must be implemented by any CDS Service that wants to be accessible from even one web-based EHR, then essentially all CDS Services will need to implement CORS.  So your suggested change makes sense -- will do.   I still see no reason to discuss the Sandbox or its CORS implementation in the spec.

PR #141, opened 2018-01-18
Reference RFC2119

bakerdb:
Added explanation re RFC2119 standards terminology.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -7,6 +7,8 @@
 
 The CDS Hooks specification describes the RESTful APIs and interactions between EHRs and CDS Services. All data in the RESTful APIs is sent and received as JSON.
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this specification are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
+
 ## Swagger (Open API Specification)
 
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -7,6 +7,9 @@
 
 The CDS Hooks specification describes the RESTful APIs and interactions between EHRs and CDS Services. All data exchanged through the RESTful APIs MUST BE sent and received as JSON structures, and MUST be transmitted over channels secured using Transport Layer Security (TLS). 
 
+### Terminology
```
brynrhodes: Suggest the term "Conformance Language" for this section, the word "Terminology" has a very different meaning in this context and is likely to be confused.

docs/specification/1.0.md
```diff
@@ -7,6 +7,9 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this specification are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
+
```
brynrhodes: Suggest using conformance language consistent with other HL7 specifications, including FHIR, as described in the [Conformance Language](http://build.fhir.org/conformance-rules.html#conflang) topic in the FHIR specification.

bakerdb: @brynrhodes  Good point.  Made change.

kpshek: I like the ability to use the full set of key words as described by rfc2119 as it is consistent with other IETF standards and gives us more verbal flexibility.

I'll discuss this with you @brynrhodes at the WGM and if there are still concerns, we can discuss further and open another issue,

PR #131, opened 2017-12-22
Clarify Card attributes, including Priority

bakerdb:
Added priority field.  
Clarified that use of GitHub Flavored Markdown, without HTML, is required.  This addresses risk #R1.

PR #143, opened 2018-01-18
CDS Service Response Changes

bakerdb:
This section was very hard to follow, especially w.r.t. what attributes were expansions of other attributes, etc.  The following changes were made:
1.  Moved the example to follow the field specifications (as we discussed).
2. Added "Priority" to all tables (except for the Card Attributes because Priorities were added by PR #131).  
3. Added sub-sections to show nesting of attributes.
4. Integrated "No Decision Support" content into "Card Array" section.
5. Integrated "Analytics" content into "Suggestion" section.

Question:  The Suggestion table references something called "Suggestion Tracking Analytics."  Is this a reference to the (moved) "Analytics" section, or to a separate resource?

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -152,6 +152,84 @@ You can see the <a href="http://editor.swagger.io/?url=https://raw.githubusercon
 
 ## CDS Service Response
 
+### Card Array 
+
+Field | Priority | Description
+----- | ----- | --------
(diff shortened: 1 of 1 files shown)


bakerdb:
@brynrhodes The original title was "Analytics," and it was placed at the end of the fields discussion.  I've inserted the title as "Suggestion Tracking Analytics" as you suggest.  Much better title, and I think better placement as well.  Also recaptured the link.  ;-)

Issue #119, opened 2017-11-21
Move CDS Service request parameters patient and encounter to Context


kpshek:
Today, the `patient` and `encounter` parameters are defined in the request to the CDS Service. However, what if we have hooks that:

- Do not have a patient in context. For instance, consider a hook `ehr-launch` that is invoked as soon as the user launches the EHR application.

- Have more than one patient in context. For instance, consider a hook `patientlist-view` that is invoked when the user is viewing their patient list.

In both of these scenarios, that current location of our `patient` and `encounter` parameters is not optimal. Either the parameters are `null` (or not set) or they are not set and instead superseded by a more appropriate parameter in the `context` in the CDS Service request.

The proposal to address this would be to remove `patient` and `encounter` from the CDS Service request and instead place these as parameters in the `context` object when appropriate for a given hook. For the three hooks we've defined and implemented today (`patient-view`, `medication-prescribe`, `order-review`), they would all include `patient` and `encounter`.

In offline discussion with @brynrhodes and @isaacvetter, we all agreed this would be a good change to make for 1.0. While this is technically a breaking change, it will serve us better down the road as we start implementing and supporting new hooks in which there is no patient or no single patient in context. Additionally, this is a minor change as we're simply just moving where in the JSON structure these parameters are read.

Given that it is a breaking change, we would appreciate thoughts from the rest of the community.

kensaku-kawamoto:
I think it's a reasonable change and a good idea.

grahamegrieve:
so to be clear, this is a significant change because you are replacing the ids of the these things with the whole resource. That's.... potentially significant. I'm making sure that this is clear

I'm in favor of this change. It works for my existing use cases, that include some where patient and encounter are not in context, and some where they are

kpshek:
> so to be clear, this is a significant change because you are replacing the ids of the these things with the whole resource. That's.... potentially significant. I'm making sure that this is clear

That is **not** what I'm proposing. Both the `patient` and `encounter` parameters would continue to just be the ids still.

brynrhodes:
> so to be clear, this is a significant change because you are replacing the ids of the these things with the whole resource. That's.... potentially significant. I'm making sure that this is clear

It would depend on the hook though, right? A hook could define the `patient` parameter as the resource, or it could define it as a `patientId` parameter. I see that as another advantage of this approach, it would be up to the hook to specify what it would actually provide.

kpshek:
[...]

PR #148, opened 2018-01-25
Calling a CDS Service Mods

bakerdb:
1) Moved the example JSON to follow the specification
2) Added priorities to the HTTP request fields table.  NOTE:  We discussed the fact that aud should not be included in the fhirAuthorization structure because fhirServer was specified outside the fhirAuthorization structure.  However, a dependency exists between the two:  if fhirAuthorization is included, a fhirServer MUST be specified.  I've made them both OPTIONAL here, and noted that if fhirAuthorization is included, fhirServer is REQUIRED.  It might be cleaner to put fhirServer inside the fhirAuthorization structure so that fhirAuthorization could be OPTIONAL, but fhirServer (as a parameter of fhirAuthorization) could be REQUIRED.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -97,6 +97,33 @@ Code | Description
 
 ## Calling a CDS Service
 
+### HTTP Request
+
+An EHR calls a CDS service by `POST`ing a JSON document to the service
+endpoint, which can be constructed from the CDS Service base URL and an
+individual service id as `{baseUrl}/cds-services/{service.id}`.  The CDS Hook
+call includes a JSON POST body with the following input fields:
+
+Field | Priority | Description
+----- | ----- | --------
+`hook` | REQUIRED | *string* or *URL*. The hook that triggered this CDS Service call<br />(todo: link to hook documentation)
+<nobr>`hookInstance`</nobr> | REQUIRED | *string*.  A UUID for this particular hook call (see more information below)
+`fhirServer` | OPTIONAL | *URL*.  The base URL EHR's [FHIR](https://www.hl7.org/fhir/) server. If fhirAuthorization is provided, this field is REQUIRED.  The scheme should be `https`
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -97,6 +97,33 @@ Code | Description
...
+
+Field | Priority | Description
+----- | ----- | --------
+`hook` | REQUIRED | *string* or *URL*. The hook that triggered this CDS Service call<br />(todo: link to hook documentation)
+<nobr>`hookInstance`</nobr> | REQUIRED | *string*.  A UUID for this particular hook call (see more information below)
+`fhirServer` | OPTIONAL | *URL*.  The base URL EHR's [FHIR](https://www.hl7.org/fhir/) server. If fhirAuthorization is provided, this field is REQUIRED.  The scheme should be `https`
+`fhirAuthorization` | OPTIONAL | *object*. A structure holding an OAuth 2.0 bearer access token granting the CDS Service access to FHIR resources, along with supplemental information relating to the token. See the [FHIR Resource Access](#fhir-resource-access) section for more information.
+`user` | REQUIRED | *string*.  The FHIR resource type + id representing the current user.<br />The type is one of: [Practitioner](https://www.hl7.org/fhir/practitioner.html), [Patient](https://www.hl7.org/fhir/patient.html), or [RelatedPerson](https://www.hl7.org/fhir/relatedperson.html).<br />For example, `Practitioner/123`
+`patient` | REQUIRED | *string*.  The FHIR `Patient.id` of the current patient in context
```
olbrich: Are you sure `patient` and `encounter` are REQUIRED?  There may be scenarios where there might not be a specific patient in context.

bakerdb: @olbrich, it did occur to me that some queries may not be patient-specific.  @kpshek, do you agree?

mattberther: Recommend holding until decision is made on #119

kpshek: @bakerdb - As @mattberther mentioned, we are going to move `patient` and `encounter` to context in #119. We are doing this because both `patient` and `encounter` are optional for hooks that are not patient/encounter specific.

In order to keep this PR moving, I'm going to merge this in but change `encounter` to `OPTIONAL`. I will have a PR for #119 later today that removes both `patient` and `encounter` out and into context for the hooks where appropriate.

PR #153, opened 2018-01-28
Move patient and encounter into hook-specific context attributes

mattberther:
This should address the outcome of the discussion in #119.


kpshek:
Thanks @mattberther! Let's get #109 reviewed/merged in first as we'll need to make some based on that new content too.

Also, I'm proposing we go with `patientId` (not `patient`) and `encounterId` (not `encounter`). This is to clearly indicate these are the FHIR ids rather than resources

bakerdb:
@kpshek It seems to me that "patient" should be OPTIONAL even when pulled within "context."  Looks like #153 makes "encounter" optional, but not "patient.

docs/hooks/medication-prescribe.md
```diff
@@ -11,11 +11,15 @@ The set of medications proposed or in progress of being prescribed. All FHIR res
 Field | Priority | Prefetch Token | Description
 ----- | -------- | ---- | ----
 `medications` | REQUIRED | No | *array* DSTU2 - An array of MedicationOrder resources <br/> *array* STU3 - An array of MedicationRequest resources
+`patientId` | REQUIRED | No | *string*.  The FHIR `Patient.id` of the current patient in context
```
jmandel: I may have missed discussion here (sorry!) -- does this imply that the patient ID can't be used as an interpolated variable in pre-fetch expressions? (I'm assuming it can... but then I'm having trouble making sense of the "No" here.)

mattberther:
@bakerdb The optional/required element as described in this PR are intended to be hook-specific. This should not be interpreted to imply that context must always include a patientId or encounterId. 

Can you think of a way we can make this clearer?

docs/hooks/medication-prescribe.md
mattberther: I dont think you missed anything @jmandel. Thanks for the second set of eyes.

I need to go back and update these three hook definitions to include the top level string elements to describe them as being able to be used as an interpolated variable.

kpshek:
> @kpshek It seems to me that "patient" should be OPTIONAL even when pulled within "context." Looks like #153 makes "encounter" optional, but not "patient.

@bakerdb - For the three hooks we have defined today, `patientId` is REQUIRED for all of them. The reason is that for each of these hooks, you always have a patient in context. However, you may not have an encounter opened. This is actually have things have always worked previously with these hooks in CDS Hooks; we're just making that explicit here.

docs/hooks/medication-prescribe.md
kpshek: Thanks for catching this @jmandel - I've fixed this on 0962406

PR #167, opened 2018-02-15
Refine authorization details around how the EHR obtains an access token on behalf of the CDS Service

### LINKED ISSUE #159: Who is talking to whom in "Obtaining an Access Token"?
I'm having trouble understanding the following text (and everything below it) in "Obtaining an Access Token":

>  If the EHR chooses to obtain a bearer token for CDS Service use, the authorization MUST be accomplished in accordance with the OAuth 2.0 authorization code grant model, and the authorization request MUST identify the CDS Service by name as the intended user. 

Who are the actors in this process? The EHR is a single entity, so I'm not clear on where/how the OAuth code flow is relevant here, and for:

> The transaction the EHR uses to request authorization includes the following parameters:

I don't know who the parties to the transaction would be.

kpshek:
This pull request removes mention of the specifics of how an EHR is to obtain an access token on behalf of the CDS Service. Instead, the documentation just lays out what the access code needs to honor.

Additionally, this pull request changes the conformance verbiage that states the access token MUST be restricted to the same permissions as the current user. Previously, this was SHOULD but after discussion with @bakerdb, @jmandel, @isaacvetter, @brettmarquard, and myself, we felt MUST was more appropriate.

This will resolve #159.


kpshek:
> should smart be uppercase (SMART) everywhere?

Yes and this pull request ensures that all references to SMART are uppercase now.

kpshek:
> Given that the fhirServer parameter (in HTTP Request) is relevant only if a fhirAuthorization is included in the request, I suggest making fhirServer a field within the fhirAuthorization parameter. This will enable us to make fhirServer REQUIRED within fhirAuthorization, thereby eliminating the need for the (current) explanation that "If fhirAuthorization is provided, this [OPTIONAL] field is REQUIRED."

@bakerdb - I had this same thought too and we discussed this on one of our previous Argonaut security calls. I had the action item to take this to the broader CDS Hooks community for feedback which uncovered a use case where it is valuable to continue keeping these two fields separated.

There is a use case in which the FHIR server is not accessible (externally) or does not offer access via the access token (in which case `fhirAuthorization` is omitted) but the data is available via prefetch; however, the `fhirServer` parameter is still useful in this scenario as the CDS Service knows where the data came from. Due to this use case (which several developers have in their production environments), keeping our fields as-is is appropriate for everyone.

bakerdb:
Thank you, @kpshek.  I knew that we had discussed this and that you were to present it to the CDS Hooks community.  But I was unaware of the resolution.  I don't fully understand why it's necessary for the CDS Service to know where EHR-prefetched data came from, but if existing production environments require the fhirServer parameter, then I see no compelling reason to move it so long as its association with fhirAuthorization is preserved.

kpshek:
Great, thanks @bakerdb. Glad we have this cleared up. 😄

Issue #125, opened 2017-12-13
Remove CORS from the specification and into a new developer guidance page


kpshek:
The CDS Hooks specification doesn't need to contain information on CORS; rather, we should move this into some new developer guidance page.

isaacvetter:
Instead of creating a "Developer Guidance" page on the website, I moved the CORS section from the spec into the existing [Security Model FAQ](https://github.com/cds-hooks/docs/wiki/Security-Model-FAQ) wiki page in PR #170

PR #170, opened 2018-03-01
address #125 - remove CORS

isaacvetter:
move CORS recommendations out of spec


isaacvetter:
Actually, don't need this change. Dixie mostly addressed in PR #140, I moved the original spec content that she had overwritten in the [wiki](https://github.com/cds-hooks/docs/wiki/Security-Model-FAQ).

Issue #154, opened 2018-01-28
Hook Versioning


AlexTatiyants:
Given that hook definitions are independent from the overall specification, is there a need to version them individually? So, for instance, one can talk about using `CDS Hooks V1.0` along with `order-review v1.2` hook.

As to what specific changes will require versioning hooks, I'm primarily concerned with changes to the hook context.

cc @kpshek

olbrich:
Perhaps the core defined hooks can be defined as part of the standard, so you can only refer to `cds-hooks v1.0 patient-view`.  Custom ones that deviate from the standard ones can use the url naming convention.

bvdh:
I totally agree, versioning hooks is a good idea. I would not make a distinction between what hooks to version and which not. Also the core-hooks can have changes.

abrooke:
A new version property can be added to the CDS Service objects returned from the  discovery endpoint, the version attribute would be located at the following JSONPath:

```
$.services[*].version
```

The version attribute would be an array of objects, where each object would indicate version compatibility of a CDS service. An example of the version compatibility object would be:

```json
{
   "cds":"1.0",
   "hook":"1.2"
}
```

The above version means that the specific CDS Service object has attributes according to CDS Hooks version 1.0 and prefetch specification compatible with hook version 1.2.  A complete CDS Services object example returned from the discovery endpoint is as follows:


```json
{
  "services": [
    {
      "hook": "patient-view",
      "title": "Static CDS Service Example",
      "description": "An example of a CDS Service that returns a static set of cards",
      "id": "static-patient-greeter",
      "prefetch": {
        "patientToGreet": "Patient/{{Patient.id}}"
      },
      "version" : [ { "cds":"1.0","hook":"1.2" } ]
    }]
}
```

The above CDS Service would be compatible with CDS Hooks version 1.0 and patient-view hook version 1.2 (with a prefetch that potentially made use of the context specification for patient-view version 1.2).

rongchen:
Versioning of individual services (hooks) are necessary. We include version information in the URL.

kpshek:
@rongchen - This issue is discussing the version of the hook definition, not the version of your CDS Service.

brynrhodes:
@bkaney, looking at #74, that's talking about services, right? This issue is talking about the version of the hook definition, not the version of the service.

bkaney:
> @bkaney, looking at #74, that's talking about services, right? This issue is talking about the version of the hook definition, not the version of the service

[...]

PR #178, opened 2018-03-15
#154: Initial draft of hook versioning support

brynrhodes:
This addresses #154

### DIFF
--- a/docs/hooks/index.md
+++ b/docs/hooks/index.md
@@ -147,6 +147,8 @@ Because hooks are such an integral part of the CDS Hooks specification, hook def
 
     specificationVersion: 1.0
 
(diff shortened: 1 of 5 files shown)


isaacvetter:
Hey Bryn,

Overall, looks good! A few potential optimizations:
1) +To help ensure the stability of CDS Hooks implementations, once a hook has been defined (i.e. published with a particular name so that it is available for implementation), it can only be changed in ways that will not break existing implementations. This means that fields can be added and restrictions relaxed, but fields cannot be changed, and restrictions cannot be tightened.

1a) "published with a particular name so that it is available" -- is this the same as an active status? If so, shouldn't we say that?

1b) "changed in ways that will not break existing " -- commonly referred to as "breaking changes" 

2)  
+Note that this means that the name of the hook carries major version semantics. That is not to say that the name must include the major version, that choice is left as a choice by users of the specification.

What would you think about adding a phrase or sentence, such as: "clean hook names increase usability. Optimally, an active hook name accurately defines the meaning and workflow of the hook in actual words".

brynrhodes:
@isaacvetter , agreed and applied

brynrhodes:
@kpshek , yep, done and committeed.

Issue #158, opened 2018-01-28
Lists of resources should be bundles


robs16:
medication-prescribe currently is using a list of FHIR resources for the medications. While this seems logically to make sense, it does not jive well with existing FHIR resource interpreting packages such as HAPI. The result is a bit of nasty code to decode as JSON, iterate the list, creating JSON strings from the resource, and then using HAPI to parse (alternatively you must do something similar but create a bundle). A list of resources is generally represented by a bundle within FHIR, and I think this is how any list of resources should be represented in cds hooks context/prefetch.

euvitudo:
One advantage of the suggestion by @robs16 is that related resources (e.g., Medication, Practitioner) could be included in the payload.  However, 'medications' would become a misnomer in this case (maybe 'prescribeBundle' or 'orderBundle'?).

kpshek:
@euvitudo - Are you referring to the `medication-prescribe` and `order-review` hooks? In that case, I think the attribute names (`medications` and `orders`, respectively) still make sense as that is the core content of those fields. By suggesting that related resources could also be included in the bundle, are you thinking of contained resources?

euvitudo:
@kpshek - mostly `medication-prescribe`, but `order-review` would be relevant too.  I'm indifferent about attribute names, though, insofar as they make sense.

Regarding resources included in the bundle--I was not thinking of contained resources (I'm honestly not a huge fan of contained resources), but bundle entries, where resources could reference other resources in the bundle so as to preclude additional retrieval of resources from a FHIR payload.

As it stands, the example in `medication-prescribe` that includes medicationCodeableConcept seems to be sufficient, but it may be useful (maybe not necessarily in `medication-prescribe`) to include other resources that are needed for service operation.

robs16:
So my service will handle this in any case, but I would generally expect medications for medication-prescribe to be a list of MedicationOrders with potentially Medication contained resources (not as bundled resources). In the case of order-review, orders may contain MedicationOrders, which in turn would also possibly contain Medications. Generally for any MedicationOrder, it is really useful for my service to have the associated Medication. There is not a good way to do this through prefetch (at least I don't think there is).  While not a 1.0 issue it might also be good to have some way to tell the EHR what we really need for context data in a future version. This could drive in the direction of #155  or what @euvitudo  is suggesting. In either case, using a bundle rather than an array works well.

PR #180, opened 2018-03-24
iss 158 - context should use Bundle, not array

isaacvetter:
switch from array of fhir medications being prescribed to bundle of fhir medications in the context for medication-prescribe hook


kpshek:
@isaacvetter - Perhaps my reading of #158 is wrong, but I thought it was saying that anytime a list of FHIR resources is used in `context`, the value should be a FHIR Bundle. The `medication-prescribe` and `order-review` hooks but have context values that fit this use case, but what about future hooks?

Do we need to add documentation/guidance stating that all lists of FHIR resources defined in future hooks also need to be valued by a Bundle?

JustinStauffer:
@kpshek - I think updating the guidance in that regard would be good.

PR #181, opened 2018-03-26
Document prefetch token format and update examples (fixes #151)

### LINKED ISSUE #151: Prefetch variables and context key/value pairs
#119 is moving some top-level fields into the context (patient, encounter). The prefetch definitions provided in discovery need to reference user, patient, encounter, and potentially other elements which may be defined in the context of the given hook. These two needs should have a clear relationship. After discussing with @kpshek, I would propose the following for prefetch definition replacements:

- Continue to use the same delimiters. ie '{{', '}}'
- Reference the user via 'user'
- Reference context entries via prefix 'context.' to the name of the context field. For example 'context.patientid'

Note that this does not support structures in the context values. Rather than provide some form of structure support (ie the id field of a MedicationOrder), define a separate field for that id in the context to make it usable in prefetch.

kpshek:
- {{User.id}} -> {{user}}
- {{Patient.id}} -> {{context.patientId}}

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -77,7 +77,7 @@ curl "https://example.com/cds-services"
       "description": "An example of a CDS Service that returns a static set of cards",
       "id": "static-patient-greeter",
       "prefetch": {
-        "patientToGreet": "Patient/{{Patient.id}}"
+        "patientToGreet": "Patient/{{context.patientId}}"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


isaacvetter:
Hey Kevin,

There was some brief conversation about this notion in PR #109. 

>+The prefetch tokens defined by this `example-hook` would be `{{context.patientId}}` and `{{context.medication.medicationId}}`.

How far down into a json object is this path syntax intended to support? Note that we're also sidestepping repeating objects. 

Why not only allow top-level strings to be referenced in context ?

Isaac

kpshek:
Thanks @isaacvetter & @brynrhodes -- Your feedback has been applied in c6c2a14

Issue #177, opened 2018-03-15
Consistent documentation of types


brynrhodes:
For tables that describe types in the spec, the name of the type is always included as the first part of the Description. These should be broken out into their own column, and we should clearly identify a list of types that are possible, starting with:

string
number
url
object
array
FHIR resource

Also, the type of the "hook" element of the Discovery and Request should be just "string", not "string or url".

kpshek:
Closed with the merge of PR #183.

Issue #173, opened 2018-03-12
Prefetch query restrictions need


brynrhodes:
The example used to illustrate prefetch query restrictions [here](http://cds-hooks.org/specification/1.0/#prefetch-query-restrictions) needs to be updated to reflect changes to context parameterization.

Specifically, it still says `patient={{Patient}}`, it should now be `patient={{context.patientId}}`.

The entire spec should be reviewed for this change.

isaacvetter:
Is this a duplicate of #151? 

Note that @kpshek addresses #151 in PR #181 and @brynrhodes addressed this issue in PR #183.

kpshek:
I agree @isaacvetter - I think this is a duplicate and has already been addressed.

Issue #172, opened 2018-03-12
Why isn't type of a context element it's own structured element?


brynrhodes:
The current description of hooks has the "type" and "description" in the same column as "description". Suggest that it would be much more useful to have "type" and "description" be separately structured.

kpshek:
Closed with the merge of PR #183.

Issue #171, opened 2018-03-12
Priority is a potentially confusing name for whether or not a context element is mandatory


brynrhodes:
The word "priority" seems problematic to describe the notion of whether or not a given context element is mandatory. Suggest "optionality", or even simply "required?" as the column header

PR #183, opened 2018-03-26
#171, #172, #173, #177: Changed Priority to Optionality in all tables…

brynrhodes:
…. Added Type column to all tables to split type out from description. Fixed several references to Patient.id (changed to context.patientId) and corrected prefetch-template documentation based on new context tokens. Also changed User.id to userId for consistency with hook context names.

### DIFF
--- a/docs/hooks/index.md
+++ b/docs/hooks/index.md
@@ -86,9 +86,7 @@ Only the `encounterId` field in this example is eligible to be a prefetch token 
 
 ## Hook Definition Format
 
-Hooks are defined in the following format:
-
------
+Hooks are defined in the following format.
 
 ### `hook-name-expressed-as-noun-verb`
 
--- a/docs/hooks/index.md
+++ b/docs/hooks/index.md
@@ -112,7 +112,13 @@ All fields defined by the hook's context MUST be defined in a table where each f
 - Type: The type of the field in the context JSON object, expressed as the JSON type, or the name of a FHIR Resource type. Valid types are *string*, *number*, *object*, *array*, or the name of a FHIR resource type. When a field can be of multiple types, type names MUST be separated by a *pipe* (`|`)
 - Description: A functional description of the context value. If this value can change according to the FHIR version in use, the description SHOULD describe the value for each supported FHIR version.
 
-While a context with FHIR data SHOULD document any differences in the data between FHIR versions, when a context object is instantiated, all FHIR resources in context MUST be based on the same FHIR version.
+### FHIR resources in context
+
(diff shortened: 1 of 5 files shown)


isaacvetter:
Hey Bryn,

It might complicate merging a bit, but I think you should also update the hooks pages to address #177. 

Isaac

isaacvetter:
Hey Bryn,

PR #181 is more targeted towards the context variable syntax change {{User.Id}} -> {{User}}.

Mind undoing those sets of changes from this PR?

Isaac

isaacvetter:
Why is so much of the diff in hooks/index.md highlighted in red?

https://github.com/cds-hooks/docs/pull/183/files#diff-64ff89a786c99c7caf0aac385c90dfb2

isaacvetter:
Is there an extra brace around the docs/hooks/medication-prescribe.md's example medications DSTU Bundle first MedicationOrder?

(This is probably my fault, not yours).

kpshek:
> Why is so much of the diff in hooks/index.md highlighted in red?

@isaacvetter - There was a missing `}` in the preceding JSON example which was causing the parse issue. I've fixed that in c956457

kpshek:
@isaacvetter - That JSON example was still a bit malformed. I fixed it and through it through a formatter (https://jsonformatter.curiousconcept.com/) in c7c8850

brynrhodes:
@isaacvetter , I agree there's probably too much content to make the same page be a template, but I like the idea of a hook template. Should I add one to this PR or should that be a separate issue?

isaacvetter:
Hey @brynrhodes ,

Undoubtedly a different issue, but perhaps one we should resolve quickly. I expect the push to create new hooks will boil over soon.

Isaac

docs/hooks/index.md
```diff
@@ -82,52 +82,59 @@ Only the `encounterId` field in this example is eligible to be a prefetch token
...
 
 Describe the set of contextual data used by this hook. Only data logically and necessarily associated with the purpose of this hook should be represented in context.
 
-All fields defined by the hook's context MUST be defined in a table where each field is described by three attributes:
+All fields defined by the hook's context MUST be defined in a table where each field is described by the following attributes:
+
+- Field: The name of the field in the context JSON object.
+- Optionality: A string value of either `REQUIRED` or `OPTIONAL`
+- Prefetch Token: A string value of either `Yes` or `No`, indicating whether this field can be tokenized in a prefetch template.
+- Type: The type of the field in the context JSON object, expressed as the JSON type, or the name of a FHIR Resource type. Valid types are *string*, *number*, *object*, *array*, or the name of a FHIR resource type. When a field can be of multiple types, type names MUST be separated by a *pipe* (`|`)
```
kpshek: What about types `true`, `false`, and `null` (all of which are allowed by JSON)?

brynrhodes: Agreed, I added *boolean* to the list. Null is allowed for the values, but I don't think we'd want the type not to be specified, right?

kpshek: Oh, good point! Yes, null shouldn't be a type 😄

PR #186, opened 2018-03-27
Document how the JWK Set is communicated (fixes #87)

### LINKED ISSUE #87: Provide guidance for where CDS Services can retrieve the EHR's public key
Open to suggestions -- but ideally we will have a few concrete proposals before discussing on a call.

kpshek:
In addition to documenting that the JWK Set is communicated either via the standard `jku` header or out-of-band, I relaxed the `iss` value from URL to URI. This is to address deployment scenarios where the issuer cannot be a URL.

Fixes #87

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -524,9 +524,10 @@ An example JSON web token header and payload:
 
 ```json
 {
-  "alg": "HS256",
+  "alg": "ES384",
   "typ": "JWT",
-  "kid": "example-kid"
+  "kid": "example-kid",
+  "jku": "https://fhir-ehr.example.com/jwk_uri"
 }
 
 {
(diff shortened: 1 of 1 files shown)


docs/specification/1.0.md
```diff
@@ -499,22 +499,27 @@ Field | Optionality | Type | Value
...
 
 The EHR MUST make its public key, expressed as a JSON Web Key (JWK) in a JWK Set, as defined by [rfc7517](https://tools.ietf.org/html/rfc7517). The `kid` value from the JWT header allows a CDS Service to identify the correct JWK in the JWK Set that can be used to verify the signature.
 
+The EHR MAY make its JWK Set available via a URL identified by the `jku` header field, as defined by [rfc7517 4.1.2](https://tools.ietf.org/html/rfc7515#page-10). If the `jku` header field is ommitted, the EHR and CDS Service SHALL communicate the JWK Set out-of-band.
```
bkaney: Consider using this link: https://tools.ietf.org/html/rfc7515#section-4.1.2

isaacvetter:
Hey @kpshek ,

A bit of feedback; overall, it's a solid design:

1) It's important that the jku value be whitelisted by the CDS Service. This was (one of) @mattvarghese's points in #87. This url should be communicated out-of-band. The CDS Service client_id creation process already exists and would be a good time to do this.

2) Our current alg example shows HS256. For commonality in implementations, the spec should dictate that CDS Services MUST support an algorithm. We're slightly partial to EC386, but HS256 would also work fine. The important thing is to standardize it.

3) Note that in the JWK RFC, kid is not actually required to be unique. Further, when it's not unique the RFC doesn't specify the actual lookup mechanism that a client should use. The CDS Hooks spec should just state that kid values MUST be unique within the jwk set. 

4) We should provide a guidance section to CDS Service developers containing the following guidelines:
  * The jku and iss urls should be whitelisted within your service. 
  * Using the ita field, CDS Services should reject any requests older than 5 minutes. 
  * CDS Services should validate that a newly received jti is unique over the past 5 minutes. This prevents replay attacks.

5) Lastly, the example json should include an example jku.

Want me to help by making commits?

Isaac

kpshek:
Thanks for the feedback @isaacvetter! I agree with all of your comments and had thought of most of these gaps. However, I was planning on addressing them via the ballot process since most of these comments are outside of the scope of this issue: how do we communicate the JWK Set in a standard way?

All of the guidance around how CDS Services should use these particular JWT header/payload fields, etc are fantastic. I just want to get this PR in so that we can cut a ballot version.

I really appreciate you helping out to address these additional concerns but given that they are outside of the scope of this issue, can we revert your additional comments here and put them on a separate branch for another issue please?

isaacvetter:
Hey Kevin,

I think that your incorrect to state that none of these changes are relevant ...

Isaac

bkaney:
> Our current alg example shows HS256. For commonality in implementations, the spec should dictate that CDS Services MUST support an algorithm. We're slightly partial to ES386, but HS256 would also work fine. The important thing is to standardize it.

I am not at all a crypto expert, but it is my understanding that HS256 is a symmetric algorithm with only one (secret) key that is shared between the two parties. ES386 is asymmetric with a public/private key pair.

After reading this: https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/ there is a catch-22 with having the algorithm be discovered in the JWT header of which is supposed to be signed. I think the upshot is the algorithm should be pre-coordinated out-of-band (or specified, as ES386). Thoughts?

isaacvetter:
Talked with @kpshek this morning:

Keeping in this PR: 
* kid should be unique
* jku example
* jku should be whitelisted

Moving to issue: 
- Guidance to CDS services to validate a JWT - how to verify a JWT? (guidance around iat, jti)
  - is this an imlpementation guide?
 - Standardizing JWT alg.

yashaskram:
jku should be a required field.

isaacvetter:
Hey Raj, 

>The EHR MAY make its JWK Set available via a URL identified by the jku header field, as defined by rfc7517 4.1.2. If the jku header field is ommitted, the EHR and CDS Service SHALL communicate the JWK Set out-of-band.

Mind creating an issue and we can talk it through as part of the ballot? We'll support jku, but I'm not convinced that you're right. 

I _think_ that right now, it's still possible for an EHR to not expose any endpoints to a CDS Hooks service and still be compliant with the spec. 

Why do you think it should be required?

Isaac

isaacvetter:
Raj! Great catch on the rfc typo - I went ahead and created #189 for this. 

I know how deeply you think about these security issues, as you think through the whitelisting considerations, they might fit well into #188. Also, your feedback is needed on #187 as well.

PR #191, opened 2018-03-30
Add copyright and CC BY 4.0 license to the footer of each page.

kpshek:
The agreement between Boston Children's Hospital and HL7 requires joint copyright and the CC BY 4.0 license on the specification.

This is in line with the agreement made for SMART.


brynrhodes:
> I used CC's own generator for this attribution

Yep, sure enough, it just looks really weird with all that blank space to the right of BY. No worries though, it's a nit for sure.

PR #193, opened 2018-03-30
Result review hook

yashaskram:
I don't exactly recollect the process for proposing new hooks. Not sure doing a PR is the right way compared to doing it formally through an issue. But I think result-review should be straightforward. This should not affect 1.0 spec finalization for upcoming HL7 ballot and so sending this your for review...

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -224,3 +224,99 @@ This example is taken from the [Opioid Prescribing Support Implementation Guide]
   ]
 }
 ```
+
+## Lab Results Intepretation 
+
+> CDS Service Request
+
+> This example illustrates the use of the CDS Hooks `result-review` hook to implement Lab Results Intepretation service.
+
+```json
+{
+  "hookInstance": "d1577c69-dfbe-44ad-ba6d-3e05e953b2ea",
+  "fhirServer": "http://fhir.example.com",
+  "hook": "result-review",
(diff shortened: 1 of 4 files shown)


kpshek:
@yashaskram - I apologize for not responding on this PR sooner. Thanks for taking the time to document and propose this new hook!

We've been mulling over the process for documenting, publicizing, and evaluating the maturity of hooks. As part of this, we intentionally separated the specific hooks with the specification as we expect there to be new hooks and changes to existing hooks far more frequently than updates to the specification.

We think that storing the hooks there in its own wiki repository will be more manageable and better reflect the relationship between CDS Hooks (the organization & specification) and the hooks defined by the community. We haven't fleshed out the details yet of the maturity model, so perhaps if a hook reaches a sufficient maturity it will be hosted on cds-hooks.org (or some other place outside of the wiki). However, we know we want to start here with the wiki.

Can you add a new Wiki page at [cds-hooks/hooks/wiki](https://github.com/cds-hooks/hooks/wiki) for your proposed `result-review` hook? Once you've done that I'll go ahead and close this PR.

yashaskram:
Ok @kpshek. I will go ahead and create a Wiki page for the result-review hook.

kpshek:
Thanks @yashaskram. Closing this.

PR #199, opened 2018-04-19
docs(examples): add hep-c-screener example service

cdaringe:
# problem

- examples list no open source services

# solution

- add our service!


brynrhodes (commented):
Love the idea, but the content is part of an HL7-managed specification and changes cannot be made without approval from the appropriate work group. Perhaps a better place for this would be the wiki, there's a [CDS Services for Testing](https://github.com/cds-hooks/docs/wiki/CDS-Services-for-testing) page in the wiki that would be a great place to post this link.

PR #414, opened 2018-10-15
update path to requirements.txt

brettmarquard:
previous instructions were to 'cd docs' but requirements.txt file is now in root. Either remove the cd or move file back to docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ $ git clone https://github.com/cds-hooks/docs
 2. Install the project dependencies
 
 ```sh
-$ cd docs
 $ pip install -r requirements.txt
 ```
 


kpshek:
@brettmarquard - The original instructions are still correct. `cd docs` refers to changing to the docs repository directory -- not docs/docs.

Here are the README commands (without the narrative explanation this may make more sense):

```sh
$ git clone https://github.com/cds-hooks/docs
$ cd docs
$ pip install -r requirements.txt
$ mkdocs serve
```

If you run those in succession from a terminal / command prompt, it all works.

brettmarquard:
well there you go, thanks!

PR #421, opened 2018-11-02
Clarify user documentation (fixes #220)

### LINKED ISSUE #220: May 2018 Ballot Comment 14
## May 2018 Ballot Comment 14

  Submitted by @bvdh from Philips Healthcare

  **Chapter:** CDS Services
  **Section:** Calling a CDS Service - http request
  **Type:** A-Q :question:
  **In Person Requested?** Yes :bust_in_silhouette:
  
  **Existing Wording:**
  > The FHIR resource type + id representing the current user.
 The type is one of: Practitioner, Patient, or RelatedPerson.
 For example, Practitioner/123
  
  **Comment:**
  What will this field mean if no FHIR server is provided? Wouldn't a full URL be better?
  
  ## Triage Information

  **Triage Notes:**
  Reviewed with Bas

  **Proposed Disposition:** Persuasive
  **Proposed Disposition Comment:**
  Change the user field documentation to state: 

The id of the user. For example, if the user represents a FHIR resource on the given FHIR server...(existing documentation we have today).
  
  ---
  _This issue was imported by @cds-hooks-bot from the [consolidated CDS Hooks May 2018 ballot spreadsheet](https://bit.ly/cds-hooks-may-2018-ballot-comments)._

kpshek:
This fixes #220


kpshek:
> Actually, this documentation was removed by the change to move user to the hook context. I'm happy to apply this requested change on top of that PR

The way we would handle this would be to merge this PR in, and then resolve the conflicts with your PR by applying this change in there.

kpshek:
@isaacvetter - Can you suggest something concrete?

Our prior documentation referenced the Patient and RelatedPerson resources -- both of which are not providers. The name of this field is `user` (or soon to be `userId`) which further reinforces the expanded intention of this field.

Should this type of clarification be best specified in each hook? For instance, medication-prescribe should probably always be a Practitioner id rather than something more broad, right?

brynrhodes:
@kpshek @isaacvetter , I agree with Kevin's suggestion that this should be specified more concretely in the hook, I'm happy to add that to the changes there.

Issue #418, opened 2018-10-29
Defining request triggers 


GitJamis:
![image](https://user-images.githubusercontent.com/20326241/47680657-11c82080-db8d-11e8-80f1-d14bdee15de6.png)

isaacvetter:
Hey @GitJamis ,

This is clearly an important issue. I really like your ![flow diagram](http://hl7.org/fhir/uv/pddi/2018Sep/assets/images/CPOE_workflow.svg).

which suggests that a CPOE system could reasonably support:
* [medication]-[select|dose|route|frequency|duration|link|sign]

Most steps for a procedure would be different, but `select` and `sign` would remain applicable. Therefore, a more reuseable set of hooks that weren't specific to medication or procedure could be `order-select` or `order-sign`. 

Two questions:
1) From an academic, clinical perspective, what would the similar set of steps be for a procedure?
2) Any concerns with my proposing `order-select` and `order-sign`and then simply identifying medication-prescribe as a medication specific subset of `order-sign` ? (Effectively, implementing Matt's suggestion from #411).

Isaac

PR #428, opened 2018-11-15
Propose order-select hook

isaacvetter:


docs/hooks/order-select.md
```diff
@@ -0,0 +1,366 @@
...
+| specificationVersion | 1.0
+| hookVersion | 0.1.0
+
+## Workflow
+
+The `order-select` hook fires when a clinician selects one or more orders to place for a patient. 
+This hook is among the first workflow events for an order entering a draft status. 
+The context of this hook may include defaulted order details 
+as it occurs immediately upon the clinician selecting the order from the order catalogue of the CPOE, 
+but before she manually specifies order details (e.g. dose, quanity, route, etc).
```
brettmarquard: update quanity to quantity

docs/hooks/order-select.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Workflow
+
+The `order-select` hook fires when a clinician selects one or more orders to place for a patient. 
+This hook is among the first workflow events for an order entering a draft status. 
+The context of this hook may include defaulted order details 
+as it occurs immediately upon the clinician selecting the order from the order catalogue of the CPOE, 
+but before she manually specifies order details (e.g. dose, quanity, route, etc). 
+Additionally, the context may include previously selected orders that are not yet signed from the same ordering session. 
+The `order-select` hook occurs after the clinician selects the order, but before she fills out (configures) the order.
```
brettmarquard: Does it make sense to add a sentence on how this is different than the 'medication-prescribe' hook? 

'This is different than the 'medication-prescribe' which is triggered after a clinician completes the details or the order.'

docs/hooks/order-select.md
```diff
@@ -0,0 +1,366 @@
+# `order-select`
+
+| Metadata | Value
+| ---- | ----
+| specificationVersion | 1.0
+| hookVersion | 0.1.0
+
+## Workflow
+
+The `order-select` hook fires when a clinician selects one or more orders to place for a patient.
```
brettmarquard: I assume this is a generic 'order-select' hook but all examples are for medications. If this applies to labs and procedures, please consider adding here.

docs/hooks/order-select.md
```diff
@@ -0,0 +1,366 @@
...
+
+Field | Optionality | Prefetch Token | Type | Description
+----- | -------- | ---- | ---- | ----
+`userId` | REQUIRED | Yes | *string* | The id of the current user.<br />For this hook, the user is expected to be of type [Practitioner](https://www.hl7.org/fhir/practitioner.html).<br />For example, `Practitioner/123`
+`patientId` | REQUIRED | Yes | *string* |  The FHIR `Patient.id` of the current patient in context
+`encounterId` | OPTIONAL | Yes | *string* |  The FHIR `Encounter.id` of the current encounter in context
+`selections` | REQUIRED | No| *array* | The FHIR id of the newly selected order(s).<br />The `selections` field references FHIR resources in the `orders` Bundle. For example, `MedicationRequest/103`.
+`orders` | REQUIRED | No | *object* | DSTU2 - FHIR Bundle of MedicationOrder, DiagnosticOrder, DeviceUseRequest, ReferralRequest, ProcedureRequest, NutritionOrder, VisionPrescription with _draft_ status <br/> STU3 - FHIR Bundle of MedicationRequest, ReferralRequest, ProcedureRequest, NutritionOrder, VisionPrescription with _draft_ status
```
kpshek: Should this be renamed as `draftOrders`?

isaacvetter: `order-review` calls this `orders`, does that change your mind?

docs/hooks/order-select.md
isaacvetter: Per conversation, this definition of the workflow step at which the `medication-prescribe` hooks fires should live in a separate issue/PR specifically about `medication-prescribe`. Further, we should consider deprecating `medication-prescribe` in favor of a `medication-select` and `medication-sign`.

docs/hooks/order-select.md
isaacvetter: We talked through this on the CDS call, there was general support for the name `draftOrders` over the current `orders`. This has been updated in `order-sign` and updating now in `order-select` PRs.

isaacvetter:
Hey @brettmarquard - will you re-review and verify that I addressed all your comments (and approve if appropriate)? @kpshek, @brynrhodes - could you please review?

docs/hooks/order-select.md
```diff
@@ -0,0 +1,366 @@
...
+| specificationVersion | 1.0
+| hookVersion | 0.1.0
+
+## Workflow
+
+The `order-select` hook fires when a clinician selects one or more orders to place for a patient, (including orders for medications, procedures, labs and other orders). 
+This hook is among the first workflow events for an order entering a draft status. 
+The context of this hook may include defaulted order details 
+as it occurs immediately upon the clinician selecting the order from the order catalogue of the CPOE, 
+but before she manually specifies order details (e.g. dose, quantity, route, etc).
```
jmandel: This distinction is unclear to me -- is it proposing that clinician-supplied details like dosing would not be passed along with this hook? I'm concerned that we're deprecating a simple model (an in-progress hook that re-fires any time an in-progress order changes) in favor of a complex model (https://github.com/cds-hooks/docs/issues/418) **but we've only introducing support for one step* of the complex model.

This strikes me as a net *reduction in functionality without simplifying* the problem.

it is possible I have misunderstood the intention of this hook, in which case it would be good to clarify here and in the definition.

isaacvetter: Hey @jmandel ,

I think that you definitely understand the intention of this hook. It fires as early as possible in the ordering workflow and will not contain clinician-specified details. It's contrasted with the `order-sign`, which fires at the last possible step in the ordering workflow and does contain all details. 

>an in-progress hook that re-fires any time an in-progress order changes

Neither [med-prescribe](https://cds-hooks.hl7.org/ballots/2018May/hooks/medication-prescribe/) nor [order-review](https://cds-hooks.hl7.org/ballots/2018May/hooks/order-review/) are documented nor, per my impression, understood nor implemented to behave in the way you're describing (frequent re-firing). 

The [Jan, 30th CDS WG call](http://wiki.hl7.org/images/7/7a/2019-01-30_CDS_WG_Call_Minutes.docx) talked through a potential `order-accept` hook which would implement an additional middle step between `order-select` and `order-sign` and represent some intermediate point at which the provider tentatively accepted order details, but before she signed.

kpshek: @jmandel / @isaacvetter - What if this hook was invoked both when then order is first selected and then also can be invoked each time the clinician specifies/selects a new detail?

FWIW, Cerner's current CPOE functionality today only triggers CDS when the order is first selected and then upon signing the order. So, exactly as how @isaacvetter has proposed here today. So, we wouldn't be able to support this in-progress hooking today.

jmandel: > Neither [med-prescribe](https://cds-hooks.hl7.org/ballots/2018May/hooks/medication-prescribe/) nor [order-review](https://cds-hooks.hl7.org/ballots/2018May/hooks/order-review/) are documented nor, per my impression, understood nor implemented to behave in the way you're describing (frequent re-firing).

FWIW, the sandbox implements them this way -- which is "code as documentation". (Seriously though, it's been the only way that developers have been able to use/test the functionality, so it's what people expect/believe about how the hook works.)

From my perspective right now, this change is introducing a more complicated system *and* reducing functionality. I can be led to reason, though! Could you help articulate the value it's introducing?

kpshek: @jmandel - What I was trying to ask/propose is what if we documented `order-select` at behaving in the way you describe but keep flexibility if a CPOE system only wanted to invoke `order-select` when the user first selects the order. Basically, I'm trying to strike a balance between the reality of some major CPOE systems today and what a re-envisioned (or newer) CPOE system should be capable of doing.

kpshek: Here are the options I see:

#### Define two hooks
We define a hook that supports selecting the order initially (eg, `order-select`).
We define another hook that supports selecting order details (eg, `order-detail-select`). Other than when the hook is invoked, the context will be identical to `order-select`.

I really don't like this option because I don't think it encourages interoperability and I don't like the duplication between the two hooks.

#### Define one hook
We define a hook that supports selecting the order initially as well as each time a detail is selected. CDS Clients are free to invoke the hook when they can.

This is my preferred option.

To support this, we could add the following to the description:

> If supported by the CDS Client, this hook may also be invoked each time the clinician selects a detail regarding the order.

Thoughts @isaacvetter on adding this additional line?

jmandel: Thanks for spelling out this proposal @kpshek. No surprise, I like it :-)

mattvarghese: Hey @jmandel, the two options Kevin spells out above and a third option of `order-cds` hook which applies at selection, modification of detail, and signing were discussed during the **HL7 Workgroup call on 01/30/2019** 
* Minutes: http://wiki.hl7.org/index.php?title=File:2019-01-30_CDS_WG_Call_Minutes.docx
* The working title of the hook in the minutes is `order-accept` where Kevin says `order-detail-select` above. 

Noteworthy excerpts:
* agree that there are 3 stages;
* EHR could implement `order-select` hook at time of accept as well
*  consensus around select and sign. Do select and sign for now; work further towards accept
* what if we had just one hook `order-cds`; this way, we are not prescriptive
* **Motion:** publish `order-select` and `order-sign` hook as currently written in github pull
request; they will have a low maturity status, and that somebody **take on a write-up of
the** `order-modify` **or** `order-accept`

jmandel: Thanks for this context @mattvarghese! When you write "EHR could implement `order-select` hook at time of accept as well," I think that's consistent with @kpshek's proposed tweaks above, to make sure that the hook documentation reflects this important fact. Does this make sense to you too?

mattvarghese: @jmandel that was not my suggestion. That was an excerpt from the minutes of the call, calling out that we discussed this (the specific line was from Ken in the minutes). See the document I linked.

In the call it was then decided this was not the way to go; rather to have separate `order-select`, `order-sign` and `order-accept` (name wasn't finalized) was the preference, after discussion of the pros and cons. 

I am personally okay with what you suggest, but I would feel like I am unfaithful to the decisions of the call if I do that. I think we should stick with the decisions of the call for now in publishing this pull request, and open separate issues to debate modifications to what was decided in the call as part of the process of maturing the hooks. Especially in this case, if we decide not to have a separate hook, but to allow `order-select` when changing the selections also, that would be a backward compatible change, so that we will not make added work for us by publishing this pull request as is.

isaacvetter: Hey Guys,

Great conversation. It's probably a good sign that this group is independently debating the same issues as CDS! These are important hooks and we're figuring out how the hook proposal process works.

With that said, I think that this level of scrutiny for the initial publishing of a low-maturity hook is generally inappropriate. A hook should be easily proposed. The in-depth feedback (and subsequent modifications and refinement) should best occur as the hook matures.

To that end, I just created issue #468 for further consideration. I'm going to resolve this conversation on this initial PR.  Are there additional obstacles to approving this PR?

Isaac

jmandel: > With that said, I think that this level of scrutiny for the initial publishing of a low-maturity hook is generally inappropriate.

I'm not sure this is fair: the "low-maturity hook" proposal is actually proposing that we deprecate a functional hook (indeed, one of only *two* functional hooks that we've been using continuously in our sandbox since September 2015). So I think the degree of scrutiny is merited, especially as this proposal would deprecate the existing functionality without replacing it.

I hope my rationale is clear here -- I'm excited to see strong discussion in this space :-)

isaacvetter: hey @jmandel ,

>this "low-maturity hook" proposal is actually proposing that we deprecate a functional hook

Well, technically it's not. We're creating the replacement for some of these hooks, but this PR doesn't deprecate either `med-prescribe` or `order-review`.

There is an existing discussion on the overall deprecation process on #433, although isn't explicitly about deprecating `med-prescribe` and `order-review`.

jmandel: I think the issue is that this PR (#428) duplicates / re-frames aspects of existing functionality, and its sibling #429 proposes to deprecate other aspects. I'm reviewing the two together as a package, since they appear to be part of the same "framework" and were considered together by the CDS Hooks workgroup, if I've understood correctly.

In any case, I don't mean to hold back progress!  My preference would be to update this PR to include the language @kpshek proposed. Perhaps this PR is a good test for our https://github.com/cds-hooks/docs/blob/master/GOVERNANCE.md process (and an opportunity to better understand how it intersects with the voting process within HL7).

jmandel: I'm hitting "unresolve" because the language @kpshek proposed isn't yet added (unless I'm missing it)

mattvarghese: @jmandel, do you know that some entity is actively relying on `medication-prescribe` being called every time some selection changes? 

1. There does not seem to be any mention in the definition of `medication-prescribe` that it was to be called every time some data changes. So I am not sure your statement stands that
>especially as this proposal would deprecate the existing functionality without replacing it.

What you are asking for replacement does not seem to be part of the `medication-prescribe` definition.

2. This was discussed extensively in the CDS Work Group call that I mentioned above, and the motion was passed 14/0/0 with no one countering or abstaining. I would think if this was actually  a problem to someone, we would have heard it by now? So is this a real world concern?

3. Are there any EMRs supporting a medication prescribe hook today that re-invokes the hook every time something is changed on the order? I would think performance considerations alone would make EHRs wary of such a choice.

jmandel: > do you know that some entity is actively relying on `medication-prescribe` being called every time some selection changes?

This has been the sandbox behavior since the beginning, so anyone who has built a hook to-date sees/expects this behavior. It's essential, for example, for any service that offers dosing advice.

mattvarghese: Well, the hook itself is badly specified is the reason it is being deprecated, right? So then, I wouldn't assign a lot of merit to the sandbox behavior - I will say, that would not be how I would expect an EHR implementation to work. 

As for dosing advice services, those should really use the proposed `order-sign` hook. It would be rather distracting, in addition to a performance personality, if it keeps popping up an alert every time the physician adjusts the dose while they haven't finalized it.

mattvarghese: PS: and that is also the argument why `order-detail-select`/`order-accept` needs to be a separate hook. While it is distracting if it keeps popping up while the physician is filling out the order, there may be some (very few) legitimate cases where it is important enough that this is still done. So the ability to distinguish between initial selection, intermediate detail change, and final signing is of paramount importance.

jmandel: We're talking in circles :-)

I obviously think the behavior is important even if it has historically been under-specified. I'm trying to prevent the behavior from be being specified *out of existence*, by instead clarifying our language in a way that *affirms the behavior*.

> if it keeps popping up an alert

In my opinion, very little or none of this advice should ever pop up an alert; that's up to each CDS client to define, of course. In any case, I think we both agree that popping up alerts in the middle of workflow makes for a frustrating UX.

To be clear, @kpshek's proposed language wouldn't require you to re-trigger the hook if you thought it would mean a performance hit within your system. It proposed an option.

Re: applying decision support at order signing, this is in my opinion a very late stage at which to be introducing changes to a dose. Much better to provide feedback early and often (and ambiently, rather than interruptively, if I can use adjectives loosely).

mattvarghese: Yes, it is a proposed option. But the option was discussed in the HL7 call, and as far as I could tell there were two conclusions
1. it was preferred to have a separate `order-detail-select` / `order-accept hook`. If we do that, then @kpshek 's proposed language will conflict with that hook. So if this need is there, a separate hook needs to be written up.
2. There did not seem like there was an urgent need for this hook during the call. At least it did not to me, and it was agreed that the `order-select` and `order-sign` hooks can be published and marked as deprecating `medication-prescribe` and `order-review` without this new hook being written up. The writing up of the new hook can be deferred till later, until good thought and use-cases can be factored into it.

mattvarghese: PS: Josh, I myself was the person who suggested in the HL7 call that we should allow order-select hook to fire at the time of accepting order edits. I was convinced by others in the call that a spearate hook is the better option. It's in the minutes:
>Matt – EHR could implement order-select hook at time of accept as well

mattvarghese: Based on my own comment in the pull request for `order-sign` https://github.com/cds-hooks/docs/pull/429#issuecomment-466241919, I am withdrawing my disagreement with adding the below to the description of `order-select` in the interest of closure.

>If supported by the CDS Client, this hook may also be invoked each time the clinician selects a detail regarding the order.

isaacvetter:
Resolves #468

docs/hooks/order-select.md
isaacvetter: Awesome!

@jmandel, when you get a chance, will you please re-review/approve and resolve, if appropriate?

Then I'll merge ... and we'll have our first proposed hook since 2015 or so! 🎉

PR #429, opened 2018-11-15
Propose order-sign hook

isaacvetter:
Note that this hook is intended to either deprecate or clarify order-review (depending upon the actual workflow intent of order-review).


docs/hooks/order-sign.md
```diff
@@ -0,0 +1,361 @@
+# `order-sign`
+
+| Metadata | Value
+| ---- | ----
+| specificationVersion | 1.0
+| hookVersion | 0.1.0
+
+## Workflow
+
+The `order-sign` hook fires when a clinician signs one or more orders for a patient.
```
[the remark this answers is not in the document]
isaacvetter: Brett - putting this out of scope, per conversation and comment on `order-select` issue.

docs/hooks/order-sign.md
```diff
@@ -0,0 +1,361 @@
+# `order-sign`
+
+| Metadata | Value
+| ---- | ----
+| specificationVersion | 1.0
+| hookVersion | 0.1.0
+
+## Workflow
+
+The `order-sign` hook fires when a clinician signs one or more orders for a patient, (including orders for medications, procedures, labs and other orders).
```
jmandel: "when a clinician signs" -- does this mean after the clinician actually clicks the "sign" button, or while a clinician is viewing the screen where the button is displayed?

From a UX perspective, it seems highly interruptive to "break" the functionality of the sign button.

isaacvetter: Hey Josh,

I wouldn't would to mandate the UI such that there was an actual button in lieu of a voice command or whatever, but yes, this hook occurs once the clinician has indicated that she has finished configuring the order and is ready to sign it. 

I think that there's a lot of existing functionality that implements cds at the point of signing. 

Isaac

jmandel: Okay, but if that's the intention, then we're missing out on a *very* important non-interruptive hook point of "I'm reviewing a pile of orders, about to sign them". If we don't address this, I'd be uncomfortable trying to deprecate our existing review hook.

kpshek: @isaacvetter / @jmandel - Does this hook need to be explicit that this is invoked when the "clinician signs" (click a button in the Cerner/Epic CPOE) vs reviewing a bunch of orders that the clinician is "about to sign"? In other words, can we support both UX paradigms?

Here is some wordsmithing:
> The `order-sign` hook fires when a clinician is ready to sign one or more orders for a patient, (including orders for medications, procedures, labs and other orders).

mattvarghese: Agree with @kpshek. I also don't think the hook should mandate the EHR workflow. 

Also, @jmandel - how is this 'order review point' different from, say if the EHR implemented `order-select` so that every time physician changes something, there will be alerts? If that is the case, I cannot programmatically distinguish the location at which `order-select` decision support needs to be made visible to physicians from the location at which `order-review` decision support needs to be made visible. A good UI would show a 'reviewable view' with the ability to change.

mattvarghese: Also, @jmandel, here we are really going in circles:
https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/Ordering.20related.20hooks.20(deprecate.20medication-prescribe)/near/153996785

jmandel: In the initial design, the distinction was basically between "I'm creating an order that's going to get added to a cart" and "I'm reviewing the set is orders in the cart". That was a meaningful distinction because it spoke to cardinality and sequence.

mattvarghese: ergo: `order-select` and `order-sign`

mattvarghese: ergo `order-select` and `order-sign`

You are yourself saying these were the original intent on one side, but on the other side you are asking that order-select should now fire everytime a dose change etc is made, and on the other side you are saying there should be another nearly redundant hook to `order-sign` which I don't see the difference from `order-sign` of.

@jmandel I really don't understand what you are wanting us to do here.

jmandel:
I am recommending that order-select be generalized to optionally include updates as the order details are entered, per @kpshek's proposal in that PR. 

And that we accept @kpshek's wordsmithed language in this PR, too.

Two small tweaks that would preserve important aspect of the current functionality.

mattvarghese:
> I am recommending that order-select be generalized to optionally include updates as the order details are entered, per @kpshek's proposal in that PR.
> 
> And that we accept @kpshek's wordsmithed language in this PR, too.
> 
> Two small tweaks that would preserve important aspects of the current functionality.

Ok fine @jmandel. I am going to merely state that 
1. I agree with the above in full and unqualified, withdrawing my disagreement on `order-select`
2. Let it be on record that I tried to uphold the decisions of the HL7 CDS Hooks workgroup meeting to the best of my ability, but failed.

Can we proceed with the hooks if these changes are made?

isaacvetter:
Hey Guys, This PR has been updated. Please review.

PR #430, opened 2018-11-15
create problem-list-item-create

isaacvetter:
Propose condition-create hook


isaacvetter:
Feedback from CDS wg: Should rename this to problem-list-create to distinguish between problist list entries and health concerns, etc.

jmandel:
Love the idea. Naming-wise, why not  call it `condition-create` as you did in the PR text? If you prefer "problem" to "condition," that's fine, but I'm confused by "list" -- since `problem-list-create` sounds like an operation to create a problem list, whereas this is an operation to create one or more *problem*(s) on a list.

jmandel:
(Similar to `appointment-book`.)

brynrhodes (approved):

isaacvetter:
CDS WG voted to merge this proposed hook into CI build today -- https://confluence.hl7.org/display/CDS/2023-06-14+-+Meeting+Agenda

vadi2 (approved):

PR #434, opened 2018-11-20
reference versioned SMART spec

isaacvetter:
Each section of the spec that references now SMART also links to the 1.0.0 version of the HL7 SMART spec.


isaacvetter:
fixes #253

PR #450, opened 2018-12-12
Require hook maturity level on hook definition

### LINKED ISSUE #444: Need to add hook maturity to the hook definition and template
Now that the hook maturity model is part of the specification, hook definitions need to include the maturity level for each hook as part of the hook metadata, and the hook definition template needs to be updated to include the default hook maturity, `0 - draft`.

isaacvetter:
Fixes #444

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -933,9 +933,9 @@ When a major change is made, the hook definition MUST be published under a new n
 > Note that the intent of this table is to outline possible breaking changes. The authors have attempted to enumerate these types of changes exhaustively, but as new types of breaking changes are identified, this list will be updated.
 
 #### Hook Maturity
-As each hook progresses through a process of being defined, tested, implemented, used in production environments and balloted, the hook's formal [maturity level](#hook-maturity-model) increases. Each hook has its own maturity level, which MUST be defined in the hook's definition and correspond to the [Hook Maturity Model](#hook-maturity-model). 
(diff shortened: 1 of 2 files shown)


isaacvetter:
I specifically left maturity changes out of the required change log, because optimally a hook should mature without substantive changes.

docs/specification/1.0.md
```diff
@@ -873,7 +873,9 @@ Hook creators SHOULD include examples of the context.
...
+The intent of the CDS Hooks Maturity Model is to attain broad community engagement and consensus, before a hook is labeled as mature, that the hook is necessary, implementable, and worthwhile to the CDS services and CDS clients that would reasonably be expected to use it. Implementer feedback should drive the maturity of new hooks. Diverse participation in open developer forums and events, such as HL7 FHIR Connectathons, is necessary to achieve significant implementer feedback. The below criteria will be evaluated with these goals in mind. The hook definition defines the hook's current maturity level and MUST link to the definition of that maturity level.
```
kpshek: > and MUST link to the definition of that maturity level.

I thought the important thing is that a hook definition MUST define its hook maturity level.

I'm not sure we should require the maturity level to be linked. Yes, linking is nice but we don't require it for anything else (like say what "prefetch token" means). I feel like that for the concept of Maturity Model, along with every other concept we have defined in CDS Hooks, requires the reader of the hook to refer back to the spec to understand everything completely.

isaacvetter:
Hey @kpshek - can you take another look at this PR? (and then #451) ?

kpshek:
Thank you @isaacvetter and sorry for re-reviewing this so late!

PR #451, opened 2018-12-12
assign maturity to patient-view, med-prescribe, order-review

### LINKED ISSUE #445: Hook definitions should be organized by maturity
The "hook catalog" should provide a clear indication of the maturity level of the hooks. One easy way to do that would be to have a folder for each maturity level, and as a hook definition progresses, it is moved through the folders.

isaacvetter:
Fixes #445 

Other than the STU ballot, I think that `patient-view` is ready for CHMM5.

Maturity Level | Maturity title | Requirements
--- | --- | ---
4 | Documented | _The above, and …_ The author agrees that the artifact is sufficiently stable to require implementer consultation for subsequent non-backward compatible changes.  The hook is implemented in the standard CDS Hooks sandbox and multiple prototype projects. The Hook specification SHALL: <ul><ol>Identify a broad set of example contexts in which the hook may be used with a minimum of three, but as many as 8-10.</ol><ol>Clearly differentiate the hook from similar hooks or other standards to help an implementer determine if the hook is correct for their scenario.</ol><ol>Explicitly document example scenarios when the hook should not be used.</ol></ul>
5 | Mature | _The above, and ..._ The hook has been implemented in production in at least two CDS clients and three independent CDS services. An HL7 working group ballots the hook and the hook has passed HL7 STU ballot.

### DIFF
--- a/docs/hooks/order-review.md
+++ b/docs/hooks/order-review.md
@@ -4,7 +4,7 @@
 | ---- | ----
 | specificationVersion | 1.0
 | hookVersion | 1.0
-| Hook maturity | [1 - Submitted](specification/1.0/#hook-maturity-model)
+| Hook maturity | [1 - Submitted](../../specification/1.0/#hook-maturity-model)
 
 ## Workflow
 
--- a/docs/hooks/order-review.md
+++ b/docs/hooks/order-review.md
@@ -4,7 +4,7 @@
 | ---- | ----
 | specificationVersion | 1.0
 | hookVersion | 1.0
-| Hook maturity | [1 - Submitted](../../specification/1.0/#hook-maturity-model)
+| Hook maturity | [3 - Considered](../../specification/1.0/#hook-maturity-model)
 
 ## Workflow
 
--- a/docs/hooks/order-review.md
+++ b/docs/hooks/order-review.md
@@ -4,6 +4,7 @@
 | ---- | ----
 | specificationVersion | 1.0
 | hookVersion | 1.0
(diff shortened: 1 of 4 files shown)


isaacvetter:
The maturity numbers next to the name of the hook is kind of ugly, but it is how [FHIR does it](https://www.hl7.org/fhir/resourcelist.html) (kind of). Maybe "patient-view [4]" or "patient-view (4)" or "patient-view : 4" ?

Any suggestions? [Heroku deployment here](https://cds-hooks-docs-pr-451.herokuapp.com/hooks/patient-view/).

image

docs/hooks/order-review.md
```diff
@@ -4,6 +4,7 @@
 | ---- | ----
 | specificationVersion | 1.0
 | hookVersion | 1.0
+| Hook maturity | [3 - Considered](../../specification/1.0/#hook-maturity-model)
```
kpshek: Unless I'm missing something, `order-review` has had far less testing than `medication-prescribe`. I don't recall anyone testing it at a Connectathon and at Cerner, we've never prototyped this hook yet. I would have marked this one as a level 2 - Tested at best.

kpshek:
@isaacvetter - I like how you have the sidebar headings today with the maturity level immediately following the hook name ("patient-view 4").

docs/hooks/order-review.md
isaacvetter: See [zulip thread](https://chat.fhir.org/#narrow/stream/17-cds-hooks/subject/how.20mature.20are.20our.20hooks.3F).

Depending how we address #429, this could be a moot point.

PR #453, opened 2018-12-21
Fix to prefetch example, for userId

### LINKED ISSUE #410: Clarification on example prefetch template
The [spec](https://cds-hooks.org/specification/1.0/#http-request_1) says that the user field in the POST request to a CDS Service would be of the format FHIR resourceType+id, in that case if the service wants the EHR to prefetch user information, the template should really be 

`prefetch: {
         user: '{{user}}'
    }`

as such the example given [here](https://cds-hooks.org/specification/1.0/#example-prefetch-templates) is wrong?

isaacvetter:
Fixes #410 

The `context.userId` field is (somewhat recently) defined to contains the FHIR resource name and identifier. For example, this field should contain: `Practitioner/123`. This was done to explicitly support scenarios other than provider-facing, for example, the user could be `Patient/456' or `Person/789` (following the [user-to-FHIR mapping](https://github.com/smart-on-fhir/smart-on-fhir.github.io/issues/148)) that SMART negotiated.

Because of this change, the prefetch example of `Practitioner/{{context.userId}}` would translate into `Practitioner/Practitioner/123` as issue #410 points out. 

This PR simply removes the "Practitioner/" string from the example, in lieu of the alternative, breaking changes suggested in #410.

### DIFF
--- a/docs/specification/1.0.md
+++ b/docs/specification/1.0.md
@@ -318,7 +318,7 @@ Only the `encounterId` field in this example is eligible to be a prefetch token 
   "prefetch": {
     "patient": "Patient/{{context.patientId}}",
     "hemoglobin-a1c": "Observation?patient={{context.patientId}}&code=4548-4&_count=1&sort:desc=date",
-    "user": "Practitioner/{{context.userId}}"
+    "user": "{{context.userId}}"
   }
 }
 ```


mattvarghese:
Aren't both the options "breaking" changes though? 
That was my thinking when I suggested instead to rename to practitionerId, since then, the resource type is implied.

mattvarghese:
Isaac explained this to me - I'm okay with going ahead with this plan.

PR #458, opened 2019-01-12
Explicitly clarify the intent behind optional and required fields

kpshek:
This came up during the Connectathon where a CDS Service implementer passed back a response in which the `links` array attribute had a `null` value. Our open source Sandbox allowed for this but other server implementers were expecting that `links` array to either be omitted or an array (empty of not).

This pull request clarifies the intent behind both optional fields as well as required string/URL fields.

While the CDS Hooks API is not defined using FHIR's rules, I did pull up how FHIR approaches this for reference:
> JSON objects and arrays are never empty, and properties never have null values (except for a special case documented below). Omit a property if it is empty

So, the proposed clarification here aligns us with what FHIR specifies (which is nice).

Note that we have just one exception to this proposed blanket rule -- a CDS Service response `cards` array. We document that this array field must always be valued, even if there is no decision support in which case the array is empty. This is (today) our one exception and the reason for the "Unless otherwise specified"

> If your CDS Service has no decision support for the user, your service should return a 200 HTTP response with an empty array of cards.

I think this is an important thing to be explicit about in our specification as it will resolve integrations challenges that will occur without this. However, we need to discuss this as a community as to whether this change is appropriate to include in our 1.0 release given how late in the release cycle we are.


isaacvetter:
Should be expanded to clarify behavior of strings, objects, arrays, including accurate terminology. Will follow-up in a later CDS meeting - 

Motion: Will incorporate language into spec to ensure that object property values are never `null`. 

Mover/Seconday: Isaac Vetter/Josh Mandel: 
Yes: 38
Abstain: 2
No: 0

kpshek:
This past Sunday (Jan 13th) we had a breakout session for CDS Hooks at the HL7 Connectathon. 30+ people we in attendance representing EHR vendors, CDS Service providers, and healthcare organizations. This PR was a topic of discussion and for the most part, the informal poll we took in the room was that the majority just want a decision on this topic (whatever that may be) and that aligning with FHIR's model in this regard was fine.

In a hallway conversation with @brynrhodes, we both think we change the only exception case we have right now (the response `cards` array which must always be valued, even with an empty array) to following this new requirement. This would mean that this line:

> If your CDS Service has no decision support for the user, your service should return a 200 HTTP response with an empty array of cards.

Would be changed to something like:

> If your CDS Service has no decision support for the user, your service should return a 200 HTTP response and the `cards` field omitted from the response object.

While that would mean the response object would be `{}` today, I foresee us adding additional top-level response fields in the future (like the system actions in the [decisions proposal](https://github.com/cds-hooks/docs/wiki/Decision-Proposal)). Did this get discussed in the working group meeting today along with the vote?

pcurc:
Is a `204/No Content` response another possibility here?

kpshek:
> Is a 204/No Content response another possibility here?

I remember thinking about this (unfortunately, there is nothing about this in our past issues). IIRC, I was thinking that we would likely add additional fields to the response which would mean that 204 likely will not be sent in all 'no decision support' scenarios.

docs/specification/1.0.md
```diff
@@ -7,6 +7,10 @@
 
 The CDS Hooks specification describes the RESTful APIs and interactions to integrate Clinical Decision Support (CDS) between CDS Clients (typically Electronic Health Record Systems (EHRs) or other health information systems) and CDS Services. All data exchanged through the RESTful APIs MUST be sent and received as [JSON](https://tools.ietf.org/html/rfc8259) structures, and MUST be transmitted over channels secured using the Hypertext Transfer Protocol (HTTP) over Transport Layer Security (TLS), also known as HTTPS and defined in [RFC2818](https://tools.ietf.org/html/rfc2818).
 
+Unless otherwise specified, JSON attributes SHALL NOT be null. If a JSON attribute is defined with an optionality of OPTIONAL, but does not have a value, implementers MUST omit it. For instance, OPTIONAL JSON string and array attributes should be omitted rather than having a null or empty value. Similarly, JSON object property values SHALL NOT be null or empty.
```
brynrhodes: I think the previous sentences cover the "property values" portion of this, so this last sentence should just say "Similarly, JSON objects SHALL NOT be empty"

PR #464, opened 2019-02-18
Propose new appointment-book hook

lmckenzi:
Proposed as part of the Davinci CRD project

### DIFF
--- a/docs/hooks/appointment-booking.md
+++ b/docs/hooks/appointment-booking.md
@@ -12,7 +12,7 @@ This hook is invoked when the user is scheduling one or more future encounters/v
 
 ## Context
 
-The Patient whose appointment is/appointments are bing booked as well as the draft Appointment records.
+The Patient whose appointment(s) are being booked, as well as the proposed Appointment records.
 
 Field | Optionality | Prefetch Token | Type | Description
 ----- | -------- | ---- | ---- | ----
--- a/docs/hooks/appointment-booking.md
+++ b/docs/hooks/appointment-booking.md
@@ -26,6 +26,7 @@ Field | Optionality | Prefetch Token | Type | Description
 
 ```json
 "context":{
+  "userId" : "PractitionerRole/A2340113",
   "patientId" : "1288992",
   "appointments" : [
     {
@@ -123,6 +124,7 @@ Field | Optionality | Prefetch Token | Type | Description
 
 ```json 
 "context":{
+  "userId" : "PractitionerRole/A2340113",
   "patientId" : "1288992",
   "encounterId" : "456",
   "appointment" : [
--- a/docs/hooks/appointment-booking.md
+++ b/docs/hooks/appointment-booking.md
@@ -4,6 +4,7 @@
 | ---- | ----
 | specificationVersion | 1.0
 | hookVersion | 1.0
+| hookMaturity | 0 - Draft 
 
 ## Workflow
 
--- a/docs/hooks/appointment-booking.md
+++ b/docs/hooks/appointment-booking.md
@@ -16,12 +16,13 @@ The Patient whose appointment(s) are being booked, as well as the proposed Appoi
 
 Field | Optionality | Prefetch Token | Type | Description
 ----- | -------- | ---- | ---- | ----
+`userId` | REQUIRED | Yes | *string* | The id of the current user.<br />For this hook, the user could be of type [Practitioner](https://www.hl7.org/fhir/practitioner.html), [PractitionerRole](https://www.hl7.org/fhir/practitionerrole.html), [Patient](https://www.hl7.org/fhir/patient.html), or [RelatedPerson](https://www.hl7.org/fhir/relatedperson.html).<br />For example, `Practitioner/123`
 `patientId` | REQUIRED | Yes | *string* | The FHIR `Patient.id` of Patient appointment(s) is/are for
 `encounterId` | OPTIONAL | Yes | *string* | The FHIR `Encounter.id` of Encounter where booking was initiated
 `appointments` | REQUIRED | No | *object* | DSTU2/STU3/R4 - FHIR Bundle of Appointments in 'proposed' state
 
 
-### Examples
+### Examples (STU3)
 
 ```json
 "context":{
@@ -120,7 +121,7 @@ Field | Optionality | Prefetch Token | Type | Description
 }
 ```
 
-```json
+```json 
 "context":{
   "patientId" : "1288992",
   "encounterId" : "456",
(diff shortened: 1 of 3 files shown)


docs/hooks/appointment-booking.md
```diff
@@ -0,0 +1,237 @@
...
+| specificationVersion | 1.0
+| hookVersion | 1.0
+
+## Workflow
+
+This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows hook services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
+
+## Context
+
+The Patient whose appointment is/appointments are bing booked as well as the draft Appointment records.
```
isaacvetter:
```suggestion
The Patient whose appointment(s) are being booked, as well as the proposed Appointment records.
```

docs/hooks/appointment-booking.md
```diff
@@ -0,0 +1,237 @@
...
+
+Field | Optionality | Prefetch Token | Type | Description
+----- | -------- | ---- | ---- | ----
```
isaacvetter: Why not include `userId` in the context (like all of our other hooks) ?

docs/hooks/appointment-booking.md
```diff
@@ -0,0 +1,237 @@
...
+          "required": "required",
+          "status": "accepted"
+        }
+      ]
+    }
+  ]
+}
+```
+
+```json
```
isaacvetter: Is this the R4 example? 

Could you please label each example with the FHIR version it implements?

docs/hooks/appointment-booking.md
```diff
@@ -0,0 +1,237 @@
+# `appointment-booking`
+
+| Metadata | Value
+| ---- | ----
+| specificationVersion | 1.0
+| hookVersion | 1.0
```
isaacvetter: Is a 1.0 version for this hook appropriate? Regardless, you should also specify a [hookMaturity](https://cds-hooks.org/hooks/template/).

```suggestion
| hookVersion | 1.0
| hookMaturity | 0 - Draft 
```

isaacvetter:
Hey Lloyd, I added a few comments in line. The only other thing you're missing is adding a menu link to the new appointment-booking hook page. I went ahead and did this for you in your branch (in the mkdocs.yml file).

Isaac

docs/hooks/appointment-booking.md
lmckenzi: Added

docs/hooks/appointment-booking.md
lmckenzi: Both are STU3.  Label added

docs/hooks/appointment-booking.md
```diff
@@ -0,0 +1,241 @@
...
+| hookMaturity | 0 - Draft 
+
+## Workflow
+
+This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows hook services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
```
kpshek:
```suggestion
This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
```

'CDS Services', not 'hook services'

jmandel:
Quick naming question, since we describe our format as noun-followed-by-verb. Our other hooks use direct verbs rather than gerunds, so I'd expect to see `appointment-book` or `appointment-reserve` or something. We hint at this in our docs, but it occurs to me we might have an opportunity to be a bit more prescriptive. Obviously not a big deal, but now's a good time to be thinking about these things.

lmckenzi:
I renamed to appointment-book as per Josh's recommendation

jmandel:
Could the description touch on why Appointment vs ServiceRequest is the target here? It may be obvious to others but based on the documentation in the core specification I'm having trouble understanding which is a better fit for the use case of layering decision support onto a scheduling or referral workflow.

lmckenzi:
@jmandel - I think that's driven by the underlying resources.  A ServiceRequest is fundamentally an authorization.  It follows the typical order workflow where you seek fulfillment (possibly directed or not), service happens, and a report comes back.  Appointment is just booking a visit.  A physician might well book an appointment to see them again, but they would never create a ServiceRequest - they don't need to authorize themselves to do something.  In fact, appointments are most typically made by the indivdidual (or at least the organization) the appointment is with.  It's *possible* to have a physician book an appointment with a different organization, but that is most commonly handled with a referral (i.e. ServiceRequest) workflow.  We have hook for ServiceRequest with the various order hooks.  The Appointment hook is for the other side of things.  If you think the documentation in the core spec is insufficiently clear, then my leaning is to raise a change request to clarify things there rather than trying to differentiate in the hook descriptions.

kpshek:
Since we have consensus on the reviews, I'm going to merge this PR in. This is still in draft state and I expect the discussion to continue around this hook.

Thanks @lmckenzi for putting this hook together!

jmandel:
@lmckenzi Just so I have the context: does this mean we'd need a different hook for managing *referrals*, like "I want my patient to see a cardiologist" (and an external service responds with "yes, you should refer to Dr. Jones, following this link")?

lmckenzi:
A referral is actually two things.  One step is the authorization process: "I, Mrs. GP have examined the patient Mr. Jones and authorize them to be seen by a cardiologist".  The second is the solicitation for care step - i.e. finding an appropriate cardiologiest to see Patient Jones.  ServiceRequest is used for the first of these.  There are a few options for the second, but most typically Task will be involved.  Separate Task instances would exist for each cardiologist asked and would reflect who said "no" and eventually who said "yes".

Actually being *seen* by the cardiologiest may involve a series of appointments - one for an initial examination, one for a follow up after some tests are done, one for a pre-surgery consult, and perhaps a couple of post-surgery follow-ups - all of which would be done under the auspices of the accepted "referral".  The initial appointment is typically arranged once the referral has been accepted by the performer.

jmandel:
(moving discussion to https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/Appointment.20scheduling)

PR #465, opened 2019-02-19
Davinci proposal for new "start of encounter" hook

lmckenzi:


isaacvetter:
Hey @lmckenzi , overall, looks great!

>Note that there can be multiple 'starts' for the same encounter as each user becomes engaged.  For example, when a scheduled encounter is presented at the beginning of the day for planning purposes, when the patient arrives, when the patient first encounters a clinician, etc.  Hooks may present different information depending on user role and Encounter.status.

Within the shoes of a CDS Client implementer, this description, while informative, is a bit loosey-goosey.

Do you think it's possible to more tightly specify the workflow or user type?  It may be the case that we're not sure yet and more implementer feedback is needed to see if a tighter spec is possible. And that's okay.

lmckenzi:
It could be relevant for multiple types of users - admissions clerk/receptionist, nurse, physician.  Each would have an interest in/influence on what tasks need to be done during the encounter and might have responsibility for following particular protocols, ensuring clinical trial requirements and/or payor requirements are met, etc.  It would be up to the service to determine what sort of information to display based on user role.  The timing of the hook is essentially when a user first deals with the encounter (or at least when they're in the early stages of dealing with the encounter).  I'm not really sure how to nail it down much more than that with general applicability, though it's possible that EHR implementers might be able to tighten it a bit more if there's more concrete consistent implementation aspects we can tie it to.

docs/hooks/encounter-start.md
```diff
@@ -0,0 +1,40 @@
+# `encounter-start`
+
+| Metadata | Value
+| ---- | ----
+| specificationVersion | 1.0
+| hookVersion | 1.0
+| hookMaturity | 0 - Draft
```
kpshek:
```suggestion
| hookMaturity | 1 - Submitted
```

You've submitted a PR and solicited feedback so you get to move up a level! 😄

docs/hooks/encounter-start.md
```diff
@@ -0,0 +1,40 @@
...
+
+Note: This is distinct from the `patient-view` hook which occurs any time the patient's record is looked at - which might be done outside the context of any encounter and will often occur during workflows that are not linked to the initiation of an encounter.
+
+The intention is that the cards from any invoked hooks are available at the time when decisions are being made about what actions are going to occur during this encounter.  For example, identifying that the patient is due for certain diagnostic tests or interventions, identifying additional information that should be collected to comply with protocols associated with clinical studies the patient is enrolled in, identifying any documentation or other requirements associated with patient insurance, etc.
```
kpshek:
```suggestion
The intention is that the cards from any invoked CDS Services are available at the time when decisions are being made about what actions are going to occur during this encounter.  For example, identifying that the patient is due for certain diagnostic tests or interventions, identifying additional information that should be collected to comply with protocols associated with clinical studies the patient is enrolled in, identifying any documentation or other requirements associated with patient insurance, etc.
```

I think you mean 'cards from any invoked CDS Services' instead of 'cards from any invoked hooks'

mkdocs.yml
```diff
@@ -21,6 +21,7 @@ nav:
     - 'order-review 3' : 'hooks/order-review.md'
     - 'order-select 1' : 'hooks/order-select.md'
     - 'order-sign 1' : 'hooks/order-sign.md'
+    - 'encounter-start 0' : 'hooks/encounter-start.md'
```
kpshek:
```suggestion
    - 'encounter-start 1' : 'hooks/encounter-start.md'
```

docs/hooks/encounter-start.md
```diff
@@ -0,0 +1,40 @@
...
+| ---- | ----
+| specificationVersion | 1.0
+| hookVersion | 1.0
+| hookMaturity | 0 - Draft 
+
+## Workflow
+
+This hook is invoked when the user is initiating a new encounter.  In an inpatient setting, this would be the time of admission.  In an outpatient/community environment, this would be the time of patient-check-in for a face-to-face or equivalent for a virtual/telephone encounter.  The Encounter should either be in one of the following states: `planned` | `arrived` | `triaged` | `in-progress`.  Note that there can be multiple 'starts' for the same encounter as each user becomes engaged.  For example, when a scheduled encounter is presented at the beginning of the day for planning purposes, when the patient arrives, when the patient first encounters a clinician, etc.  Hooks may present different information depending on user role and Encounter.status.
+
+Note: This is distinct from the `patient-view` hook which occurs any time the patient's record is looked at - which might be done outside the context of any encounter and will often occur during workflows that are not linked to the initiation of an encounter.
```
jmandel: This is super helpful in terms of understanding the thought process here. For service developers, do we have examples/thoughts about when they'd want to support *only* `encoutner-start-` and *not* also support `patient-view`? May recommendations that make sense in the context of deciding what to do during the current encounter might also be reasonable in the context of planning for a future encounter.

lmckenzi: If a pre-authorization is needed for a service, it makes sense to show it at time of admission, but not every time a nurse or physician looks at a record during a week-long stay.  The same would be true if there are diagnostic tests the patient is due for or steps of a clinical trial that need to be done during a scheduled encounter.  The primary purpose of the separation is to help keep hooks from being overwhelming or noisy.  If the intention is "show them once, let them make a decision and don't bug them again", enounter-start is appropriate.  If the intention is keep bugging the clinician even if the patient is no longer in their care until a record is updated, then patient-view makes sense.

jmandel: This generally makes sense, and is good flavor.

If you've got a pre-authjorization step required and someone *doesn't* actually complete it at admission time, do you want it to just disappear from the virtual todo list?

Also would note that a given CDS Service can track which users it has prompted, etc, and could decide to stop showing cards.

Anyway, I'm not challenging the utility of this hook; I just want to do as much as we can to help articulate when each hook is more appropriate, and when a developer would want to support both.

kpshek:
Since we have consensus on the reviews, I'm going to merge this PR in. This is still in draft state and I expect the discussion to continue around this hook.

Thanks @lmckenzi for putting this hook together!

PR #469, opened 2019-02-26
Deprecate medication-prescribe and order-review

mattvarghese:
Pull request to deprecate `medication-prescribe` and `order-review` hooks now that `order-select` and `order-sign` hooks have been created.

### DIFF
--- a/docs/hooks/order-review.md
+++ b/docs/hooks/order-review.md
@@ -8,7 +8,7 @@
 
 ## Deprecation Notice
 
-This hooks is being deprecated in favor of the `order-sign` hook. The workflow during which this hook fires is unclear and has led to varying implementations and confusion. Therefore, after discussions on Zulip, and the HL7 CDS Committee, the `medication-prescribe` and `order-review` hooks are being deprecated in favor of newly created [`order-select`](../order-select) and [`order-sign`](../order-sign) hooks. This notice is a placeholder to this effect while CDS Hooks determines the [appropriate process for deprecating hooks](https://github.com/cds-hooks/docs/issues/433).
+This hooks is being deprecated in favor of the `order-sign` hook. The workflow during which this hook fires is unclear and has led to varying implementations and confusion. Therefore, after discussions on Zulip, and the HL7 CDS Committee, the `medication-prescribe` and `order-review` hooks are being deprecated in favor of newly created [`order-select`](../order-select) and [`order-sign`](../order-sign) hooks. Native criteria within the EHR will be relied upon to only fire the new hooks for appropriate medications / procedures. This notice is a placeholder to this effect while CDS Hooks determines the [appropriate process for deprecating hooks](https://github.com/cds-hooks/docs/issues/433).
 
 ## Workflow
 
--- a/docs/hooks/order-review.md
+++ b/docs/hooks/order-review.md
@@ -8,7 +8,7 @@
 
 ## Deprecation Notice
 
-This hooks is being deprecated in favor of the `order-sign` hook. The workflow during which this hook fires is unclear and has led to varying implementations and confusion. Therefore, after discussions on Zulip, and the HL7 CDS Committee, the `medication-prescribe` and `order-review` hooks are being deprecated in favor of newly created [`order-select`](../order-select) and [`order-sign`](../order-sign) hooks. Native criteria within the EHR will be relied upon to only fire the new hooks for appropriate medications / procedures. This notice is a placeholder to this effect while CDS Hooks determines the [appropriate process for deprecating hooks](https://github.com/cds-hooks/docs/issues/433).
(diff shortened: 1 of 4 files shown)


mattvarghese:
@isaacvetter, @jmandel, @kpshek, @brynrhodes, @brettmarquard 
Hey Folks, I've made this pull request to deprecate the old hooks. Could you folks review this? Also, I'm unable to officially request review on the right side reviewers list (probably since I'm not a committer). Can one of you request review for me?

jmandel:
Thanks @mattvarghese!

In our discussion, I thought the proposal was to deprecate the order-review hook, but not to deprecate medication-prescribe. To be clear, I'm okay with deprecating both -- I'm just trying to reconcile what we discussed with the changes here to make sure I understand.

mattvarghese:
@jmandel, the proposal in all discussions I have been involved have been to deprecate both. 

1. Especially, `medication-prescribe` has been interpreted by some people to mean the `order-select` workflow point, and others to mean the `order-sign` workflow point, so that there is even more confusion with `medication-prescribe` than with `order-review`. 

2. Plus, `medication-prescribe` has the limitation of being medication specific, whereas the workflow points in all EHRs we have discussed allow ordering both medications and procedures together, and the ability to provide decision support on procedures is as much valuable as on medications (if anything, there are more live web-service based procedure decision support using EHR specific equivalents / one off integrations than there are for medications.)

And the CDS Workgroup call on this topic had also come to the same agreement on deprecating both in favor of the new hooks. 

3. Looking at it from the opposite angle, if `order-select` and `order-sign` exists, is there any workflow previously attributed to `medication-prescribe` that is not covered by them? 

The consideration that `medication-prescribe` only contains medications was also discussed at the HL7 workgroup call. 

4. It is really an large burden on the EHR to support different hooks based solely on the endpoint on the other side, where workflow point remains the same. Whereas it is relatively easy for the recipient to ignore ProcedureRequests if it doesn't care about them.  (There is also the argument that medication decision support should not be made in isolation, without the awareness of procedures being ordered - patient safety.) So along with the deprecation, I would also request that we avoid conflict of workflow points across hooks where possible - right now `order-select` and `order-sign` conflict with `medication-prescribe`.

5. The reason why many people wanted `medication-prescribe` was that they only care about the medications. But all EHRs already allow native criteria to prevent the hook being called if there is no medication / specific medication, that the endpoint cares about, being ordered. So the larger value comes from specifying in the EHR that this hook needs to be called only for say blood thinners, rather than sending all medications to a CDSS whenever a medication is ordered and have it return no cards unless a blood thinner is present (reduces the number of web service calls and therefore the load on the CDSS *substantially* compared to using`medication-prescribe`.) Therefore, the targeting of decision support can actually be done much better by having generic hooks, and leveraging native criteria in the EHR as a 'pre-check'. If we go with `medication-prescribe` instead, why would we stop at that rather than say `bloodthinner-prescribe` (again, I am paraphrasing from the discussion we had on the HL7 call in this fourth point). 

6. There have been discussions on making a framework for CDS Services to specify the *targeting criteria* currently accommodated through EHR native abilities in 5 above directly in the CDS Hooks specification. I support this. However, considering the size of the scope of that discussion, I would request to keep that separate from this.

jmandel:
Thank -- the thing I was confused about is that #428 didn't mention deprecation while #429 did. I'm happy with deprecating both.

mattvarghese:
@jmandel, yes, I noticed that, and have updated both `order-select` and `order-sign` to call this out, in addition to the deprecation notices on `order-review` and `medication-prescribe`. 

Hope that will be good clarification for the interim until we decide on a proper deprecation process? 

My primary intent with this pull-request is that new use-cases don't start with `order-review` or `medication-prescribe` and old use-cases will have a longer notice period than if we wait till we agree on a proper deprecation process.

mattvarghese:
Also, thanks for adding the reviewers for me!

docs/hooks/order-review.md
```diff
@@ -6,6 +6,10 @@
 | hookVersion | 1.0
 | Hook maturity | [3 - Considered](../../specification/1.0/#hook-maturity-model)
 
+## Deprecation Notice
+
+This hooks is being deprecated in favor of the `order-sign` hook. The workflow during which this hook fires is unclear and has led to varying implementations and confusion. Therefore, after discussions on Zulip, and the HL7 CDS Committee, the `medication-prescribe` and `order-review` hooks are being deprecated in favor of newly created [`order-select`](../order-select) and [`order-sign`](../order-sign) hooks. Native criteria within the EHR will be relied upon to only fire the new hooks for appropriate medications / procedures. This notice is a placeholder to this effect while CDS Hooks determines the [appropriate process for deprecating hooks](https://github.com/cds-hooks/docs/issues/433).
```
jmandel:
```suggestion
This hook is deprecated in favor of the `order-sign` hooks, with the goal of  clarifying workflow trigger points and supporting orders beyond medications. In this refactoring, `medication-prescribe` and `order-review` hooks are being deprecated in favor of newly created [`order-select`](../order-select) and [`order-sign`](../order-sign) hooks. This notice is a placeholder to this effect while CDS Hooks determines the [appropriate process for deprecating hooks](https://github.com/cds-hooks/docs/issues/433).
```

PR #471, opened 2019-03-07
Mark appointment-book, encounter-discharge and encounter-start as mat…

lmckenzi:
…urity 1 as they've now been posted in the registry and feedback has been solicited on http://chat.fhir.org as per: 
https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/Appointment-book
https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/Encounter-start
https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/Encounter-discharge


lmckenzi:
Note: I didn't change the 'version' of the hooks because I presumed that version changes are only necessary if there's a substantive (or at least implementer-relevant) change to the hook content.  I'm assuming that changes of level (or even minor formatting/typo corrections) would not require a change to the hook version.

PR #473, opened 2019-03-08
Further clarify the purpose of the 'selections' context

lmckenzi:
It wasn't totally clear how or whether the CDS service was actually supposed to behave differently towards orders that were 'selected' vs. not.  The proposed wording is my guess at the intent, but happy to have the proposed wording changed to something else so long as it clarifies what the point is of differentiating selected from non-selected orders in terms of the behavior of the service.


docs/hooks/order-select.md
```diff
@@ -19,7 +19,9 @@ This hook is intended to replace (deprecate) the `medication-prescribe` hook.
 
 ## Context
 
-The context of this hook distinguishes between the list of unsigned orders from the clinician's ordering session, and one or orders just added to this list. The `selections` field contains a list of ids of these newly selected orders; the `draftOrders` Bundle contains an entry for all unsigned orders from this session, including newly selected orders.
+The context of this hook distinguishes between the list of unsigned orders from the clinician's ordering session, and the one or two orders just added to this list. The `selections` field contains a list of ids of these newly selected orders; the `draftOrders` Bundle contains an entry for all unsigned orders from this session, including newly selected orders.
+
+Decision support should focus on the 'selected' orders - those that are newly selected or actively being authored.  The non-selected orders are included in the context to provide context and to allow decision support to take into account other pending actions that might not yet be stored in the system (and therefore not queryable).
```
brynrhodes: The wording here implies that selected orders _are_ stored, but I don't think that's necessarily justified. Suggest dropping the end of the last sentence " that might not yet..."

isaacvetter: @lmckenzi - mind approving Bryn's suggested changes?

PR #479, opened 2019-05-18
Add logo-use page linking to SMART guidelines

jmandel:
This is a step toward resolving #479. The linked content at smarthealthit.org is not yet available, but is a work in progress. It would be good to incorporate this general guidance into the official docs, linking out to the more detailed guidelines.


docs/logo-use.md
```diff
@@ -0,0 +1,12 @@
+# Use of CDS Hooks Logos
+
+<center>
+<img src="/images/logo.png">
+<img width="60" src="/images/logo-square.png">
+</center>
+
+The CDS Hooks logos are owned by The Children's Medical Center Corporation and may be used in accordance with guidelines and policy at [https://smarthealthit.org/cds-hooks-logo](https://smarthealthit.org/cds-hooks-logo). These guidelines describe how CDS Hooks logos can be used in connection with:
```
jmandel: Note that this is a 404 today; SMART will be adding content similar to https://smarthealthit.org/logo/ (but tailored for CDS Hooks, and without the need for a distinct "works with" logo.)

isaacvetter:
Wouldn't it make more sense to wait until https://smarthealthit.org/cds-hooks-logo exists before merging? Why put a broken link on cds-hooks.org? Broken links simply degrade a visitor's trust in the site.

jmandel:
We're working on getting content up, indeed. BCH counsel is drafting an updated, simplified set of guidelines. I think having the notice is strictly better than not -- but we can commit to getting a placeholder up to avoid a broken link ASAP, even before the guidelines are compete.

isaacvetter:
Hey Josh,

First - great work getting progress from BCH counsel! This is important stuff.

I recommend three changes:
1) Don't publish until there's actually an accurate legal description of how the logo can be used. Otherwise, the cds-hooks.org page specifically does not contain allowed uses of the logo.
2) Move the "Logo Use" link from the main menu into the footer with the rest of our legalese. 
3) Since the cds-hooks.org/logo-use page is explanatory text and not the actual legal use license, I think that we should soften the "[The Children's Medical Center Corporation](https://www.google.com/search?q="the+children%27s+medical+center+corporation"&oq="the+children%27s+medical+center+corporation")" term, which is not a recognizable name and sounds extremely corporate. Preferably, we'd simply call it the "Boston Children's Hospital" (like we do in the footer). If this isn't legally advisable, I'd suggest explaining, like:
>The CDS Hooks logos are owned by The Children's Medical Center Corporation (the parent company of Boston Children's Hospital) ...

Apologies for the tardiness of this feedback.

Isaac

jmandel:
To be clear, I merged this after @brynrhodes's review yesterday. I'm happy to move the link to the bottom of the page, and to update the language to read "Boston Children's Hospital" per your suggestion (will do this in a follow-on PR). It's very helpful though to have this content in place as we get detailed guidance onto the smarthealthit.org site, so I hope it's okay that we have a pointer to content-in-progress. (To be fair, the terms are always in-progress in the sense that they can be changed, so I don't see this as a huge limitation, and in any case we should have more specifics shortly.)

jmandel:
Updated https://github.com/cds-hooks/docs/pull/480 with your suggestions @isaacvetter

isaacvetter:
Is https://smarthealthit.org/cds-hooks-logo/ complete?!?!

jmandel:
I think so -- is there something surprising about it (given the double interrobang)??

PR #480, opened 2019-05-20
Move logo use link

jmandel:


jmandel:
Note that links in this portion of the footer aren't very accessible: they render with no visible indication:
image

I'd suggest we update the css to make it clear this logo language is a link (e.g., with an underline or a font color change).

jmandel:
My take: moving this information to the footer makes it harder to find and use. But I'm fine with this change if others think it's an improvement

FYI @isaacvetter @brynrhodes

isaacvetter:
Josh, my take is that this information should be a little harder to find. This is meta content that's not the purpose of the site. @brynrhodes , @brettmarquard - Josh and I disagree (never a good sign and usually means that I'm wrong), what'd'ya think?

Totally agree with the lack of a link overview and lack of different font color.

brynrhodes:
@jmandel , @isaacvetter , I don't feel strongly about this, but if I look at the FHIR site, the footer just has the creative commons image and that links to a "License" page in the FHIR spec that discusses licensing and trademark issues. I wonder if a similar approach is appropriate here, add a License page, linked from the Creative Commons logo in the footer, and that would be where the logo use guidelines link would be, along with general licensing information about the spec. Having said that, I agree with Isaac that it's the kind of thing that makes sense to have in the footer, so if we don't want to take the extra step of putting a full license page in, then I'm good with this PR.

jmandel:
I think overall licensing language would be great to have -- and it would be nice to consolidate.  on the licensing front, do we have anything to say beyond naming and linking to the creative commons license for the spec, and mentioning that the sandbox and other code are licensed as Apache?

brynrhodes:
Perhaps a disclaimer, similar to the one in [FHIR](http://hl7.org/fhir/license.html#disclaimer)?

isaacvetter:
Per conversation wih @brynrhodes , @brettmarquard , we should start with a simple page that adds a disclaimer and the link to BCH's logo use page.

The disclaimer should be something like FHIR's, per Bryn:

> Disclaimer and Warning of Use 
>CDS Hooks is derived from the considerable collective experience of the HL7 membership and wide community feedback from the development and application of a spectrum of health care interoperability solutions. However, CDS Hooks is generalized to support multiple contexts of use. It is the responsibility of the persons or organizations using this specification to ensure its use is fit for the particular purpose in which it is used, including validation for clinical and operational use.
>
>See also the specific warnings associated with use of the STU.

jmandel:
Sounds good! Should we close this PR and open a separate one to address?

isaacvetter:
@jmandel - what do you think of #482?

jmandel:
I think it's on the right track! Will close this one.

PR #484, opened 2019-06-19
Add logo vector images + TM symbol

jmandel:
This adds `svg` source logos, plus TM symbol, which is one step along the way to adding a registered trademark.


jmandel:
Thanks! Will merge and take this as follow-up feedback.

PR #497, opened 2019-09-16
Add preference indication to Suggestion

mattvarghese:
Allows a service to indication how to preferable this suggestion is from among all the available suggestions on the card.


mattvarghese:
chat.fhir.org discussion: https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/CDS.20Hooks.20feature.20request

mattvarghese:
@jmandel , @isaacvetter , @brynrhodes , @dennispatterson - requesting reviews :)

docs/specification/1.0.md
```diff
@@ -473,6 +473,7 @@ Field | Optionality | Type | Description
 ----- | ----- | ----- | --------
 `label` | REQUIRED | *string* | Human-readable label to display for this suggestion (e.g. the CDS Client might render this as the text on a button tied to this suggestion).
 `uuid` | OPTIONAL | *string* | Unique identifier, used for auditing and logging suggestions.
+`preference` | OPTIONAL | *string* | Allows a service to indicate how preferable this suggestion is from among all the available suggestions on the card. Allowed values are: `recommended`, and `alternate`. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting recommended suggestions etc.
```
jmandel: Thanks for tagging me here @mattvarghese !

I like this concept and I'm trying to wrap my head around the model. For instance, why not just add an `isPreferred` boolean field? A few questions to explore in this vein...

1. Is the intention to add more values to the `{recommended, alternate}` enum over time? 
2. Is it allowable to mark more than one suggestion as `recommended`?
3. Is it allowable to mark all suggestions as `alternate`?

isaacvetter:
Hey @mattvarghese , @jmandel ,

When you get a chance, will you look at my proposed changes, which also may coincide nicely with answering the above questions?

What do ya'll think?

Isaac

jmandel:
> will you look at my proposed changes, which also may coincide nicely with answering the above questions?

 @isaacvetter I'm missing something -- where are the proposed changes/

isaacvetter:
@jmandel - I think that I committed them directly to Matt's branch, I can see them here: https://github.com/mattvarghese/docs/commit/675b078c07124cab819edb75606d4e376a2159f7#diff-7971a89e057af84349a92f9ee9064c66


And they consist of this:

>`isRecommended` | OPTIONAL | *boolean* | Allows a service to indicate that a specific suggestion is recommended from all the available suggestions on the card. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting, or highlighting recommended suggestions. Multiple suggestions MAY be recommended, if `card.selectionBehavior` is `any`. A CDS Service MUST only recommend one (or more) suggestions if that service expects the suggestion(s) to be accepted a majority of the time.

jmandel:
Thanks -- I see your changes now @isaacvetter. I like this model; I'll propose a quick line edit here in the PR...

docs/specification/1.0.md
```diff
@@ -473,6 +473,7 @@ Field | Optionality | Type | Description
 ----- | ----- | ----- | --------
 `label` | REQUIRED | *string* | Human-readable label to display for this suggestion (e.g. the CDS Client might render this as the text on a button tied to this suggestion).
 `uuid` | OPTIONAL | *string* | Unique identifier, used for auditing and logging suggestions.
+`isRecommended` | OPTIONAL | *boolean* | Allows a service to indicate that a specific suggestion is recommended from all the available suggestions on the card. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting, or highlighting recommended suggestions. Multiple suggestions MAY be recommended, if `card.selectionBehavior` is `any`. A CDS Service MUST only recommend one (or more) suggestions if that service expects the suggestion(s) to be accepted a majority of the time.
```
```suggestion
`isRecommended` | OPTIONAL | *boolean* | Allows a service to indicate that a specific suggestion is recommended from all the available suggestions on the card. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting, or highlighting recommended suggestions. Multiple suggestions MAY be recommended, if `card.selectionBehavior` is `any`. 
```

I'd recommend cutting out the following sentence: 

> A CDS Service nd one (or more) suggestions if that service expects the suggestion(s) to be accepted a majority of the time.

... because I'm not sure it's the right rubric, and I don't know how a service should derive "expectations." In other words, the combination of hard conformance language ("MUST") with a squishy assessment doesn't sit quite right with me.

dennispatterson:
Would it be better to incorporate a ranking mechanism rather than limiting to say that one suggestion is recommended?

brynrhodes:
Using a Boolean aligns with [PlanDefinition.precheckBehavior](http://build.fhir.org/plandefinition-definitions.html#PlanDefinition.action.precheckBehavior)

dennispatterson:
Thanks, @brynrhodes!  Discussed on the call with you and @isaacvetter that this is more about UI and providing a default.

dennispatterson:
I can see a CDS Service wondering if they need to always return this.  Thoughts on guidance saying that this should only be populated when there's more than one suggestion or something along those lines?

mattvarghese:
@dennispatterson - I can see a CDS Service saying something like "We suspect the patient has XYZ. If you agree, do ABC" - maybe 40% confidence. In this case, maybe ABC should not be recommended in the card. 
Whereas if it was, "Patient is overdue for a XYZ exam. Please order one" has a confidence of 90+%. So maybe that should come recommended on the card?

isaacvetter:
Hey @dennispatterson ,

Building on Matt's response -- 

I had a similar concern: what guidance should the spec give to CDS developers to help them determine when to set `isRecommended` to true? Note that I'd initially even included this sentence as part of this PR:

> A CDS Service MUST only recommend one (or more) suggestions if that service expects the suggestion(s) to be accepted a majority of the time.

This loose definition of "a majority of the time" is also what Matt's describing in his contrast of 40% vs 90% of the time.  [Josh's comment](https://github.com/cds-hooks/docs/pull/497/files/f12551407f3dac4506481875053071b7727ca8f8) 

> not sure it's the right rubric, and I don't know how a service should derive "expectations."

is why I removed this sentence.

Teasing apart his comment, here's the rationale for not providing this guidance: 
1) Especially without the feedback service -- how does a service actually know that a suggestion is the more commonly chosen?

1) More importantly -- Recommending a suggestion because it's _usually_ the right choice ("majority of the time") may not actually be a _safe_ guideline. It's easy to imagine cases where the typical choice could be unsafe for a given patient and clinical judgement is required. 

These are the reasons why is it's likely better for the spec to not yet provide guidance to developers.

Is this convincing?

Isaac

p.s. 
> only be populated when there's more than one suggestion or something along those lines?

Dennis, also note that it could make sense to populate `isRecommended` when there's only one suggestion. Matt's example: patient is overdue for exam. The "order exam" suggestion is the only suggestion, it's low risk, we want to encourage the provider to follow some care path/best practice, it would make sense to set `isRecommended`, even if it were the only suggestion.

dennispatterson:
Fully agree on not giving arbitrary guidance around confidence levels or what's "majority."

My hangup is more understanding the value in a CDS Service indicating this when there's only one suggestion.  By default, a single suggestion would be recommended, right?  If the single suggestion is "order exam", why is an extra indicator needed to denote that the suggestion is recommended?  Isn't it recommended by virtue of being returned?

I'm also trying to steer away from CDS Services being pressured to always return this on single suggestions so that it holds more weight or isn't _not_ recommended.

docs/specification/1.0.md
```diff
@@ -473,6 +473,7 @@ Field | Optionality | Type | Description
 ----- | ----- | ----- | --------
 `label` | REQUIRED | *string* | Human-readable label to display for this suggestion (e.g. the CDS Client might render this as the text on a button tied to this suggestion).
 `uuid` | OPTIONAL | *string* | Unique identifier, used for auditing and logging suggestions.
+`isRecommended` | OPTIONAL | *boolean* | Allows a service to indicate that a specific suggestion is recommended from all the available suggestions on the card. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting, or highlighting recommended suggestions. Multiple suggestions MAY be recommended, if `card.selectionBehavior` is `any`.
```
dennispatterson:
```suggestion
`isRecommended` | OPTIONAL | *boolean* | When there are multiple suggestions, allows a service to indicate that a specific suggestion is recommended from all the available suggestions on the card. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting, or highlighting recommended suggestions. Multiple suggestions MAY be recommended, if `card.selectionBehavior` is `any`.
```

isaacvetter: added manually

PR #498, opened 2019-09-16
Add selectionBehavior option of 'any'

mattvarghese:
Added "any" and "all" as options for selectionBehavior. Also cleaned up to remove redundant language.


alexander-ivanov:
We were asked for some use-cases for 'all' and 'any' types of selection behavior. We use cds-hooks with own type of cds-hooks, this hook returns several independent actions which can be performed. We can encode them as a permutation of all actions, each permutation as separate suggestions, and use selection behavior - 'at-most-one' but it produces very large output and hurts performance. So we would like to encode each action as a separate suggestion and give the user the possibility to take several of them.

@Sergey778 may want to add some comments.

docs/specification/1.0.md
```diff
@@ -452,7 +452,7 @@ Field | Optionality | Type | Description
...
-`selectionBehavior` | OPTIONAL | *string* | Describes the intended selection behavior of the suggestions in the card. Currently, the only allowed value is `at-most-one`, indicating that the user may choose none or at most one of the suggestions. Future versions of the specification may expand this behavior, so CDS Clients that do not understand the value MUST treat the card as an error. CDS Clients MUST support the value of `at-most-one`.
+`selectionBehavior` | OPTIONAL | *string* | Describes the intended selection behavior of the suggestions in the card. Allowed values are: `at-most-one`, indicating that the user may choose none or at most one of the suggestions, `any`, indicating that the end user may choose any number of suggestions including none of them and all of them, and `all`, indicating that all suggestions MUST be chosen when accepting the card. CDS Clients that do not understand the value MUST treat the card as an error.
```
```suggestion
`selectionBehavior` | OPTIONAL | *string* | Describes the intended selection behavior of the suggestions in the card. Allowed values are: `at-most-one`, indicating that the user may choose none or at most one of the suggestions;`any`, indicating that the end user may choose any number of suggestions including none of them and all of them. CDS Clients that do not understand the value MUST treat the card as an error.
```

The desired "all" behavior can be achieved with existing functionality, by listing multiple `actions` in a single suggestion's actions array. @mattvarghese can you review and merge this tweak if you agree?

mattvarghese:
I've updated the PR per Josh's suggested change, to only include 'at-most-one' and 'any', and removed 'all' option.

mattvarghese:
@isaacvetter , @brynrhodes , @dennispatterson - requesting reviews :)

docs/specification/1.0.md
```diff
@@ -452,7 +452,7 @@ Field | Optionality | Type | Description
...
-`selectionBehavior` | OPTIONAL | *string* | Describes the intended selection behavior of the suggestions in the card. Currently, the only allowed value is `at-most-one`, indicating that the user may choose none or at most one of the suggestions. Future versions of the specification may expand this behavior, so CDS Clients that do not understand the value MUST treat the card as an error. CDS Clients MUST support the value of `at-most-one`.
+`selectionBehavior` | OPTIONAL | *string* | Describes the intended selection behavior of the suggestions in the card. Allowed values are: `at-most-one`, indicating that the user may choose none or at most one of the suggestions;`any`, indicating that the end user may choose any number of suggestions including none of them and all of them. CDS Clients that do not understand the value MUST treat the card as an error.
```
dennispatterson: Any concerns with removing a 1.0-spec requirement that 'at-most-one' be supported?  I understand it was required as being the only option, but trying to think through any potential issues with removing something that was stated as a requirement.

mattvarghese: I'm in fine with the above wording. Do I need to do anything now - I don't see an "accept changes" button or anything.

isaacvetter: Hey guys,

`selectionBehavior` is OPTIONAL. Making is REQUIRED would be a breaking change. If no selectionBehavior is provided, how does a client know what behavior to enact?

Isaac

dennispatterson: Our call on 11/13 walked through my initial comment and I think we clarified.

Not sure I understand your comment, Isaac, on the optionality of selectionBehavior and "If no selectionBehavior is provided, how does a client know what behavior to enact?"  I know you're familiar with the [conditionally required](https://github.com/cds-hooks/docs/issues/509) aspect of selectionBehavior, so can you clarify?

PR #496, opened 2019-09-16
Add uuid to card

mattvarghese:
Allow cards to have a card identifier.

PR #505, opened 2019-10-09
card.source.topic: allow CDS Service to specify a topic for high-level categorization

mattvarghese:
Adding a ruleCode field to Card.source, of type FHIR Coding. This allows the CDS Service to expose metadata about the rule it used to generate the Card. Knowledge of this information can allow the CDS Client to be more nuanced in its presentation of the Card. 

Discussion: https://chat.fhir.org/#narrow/stream/179159-cds-hooks/topic/CDS.20Hooks.20feature.20request


mattvarghese:
@jmandel , @isaacvetter , @brynrhodes , @dennispatterson - requesting reviews :)

bvdh:
See review comment https://github.com/cds-hooks/docs/pull/505#pullrequestreview-299885651

mattvarghese:
@bvdh 

I can see the point behind `adviceCode` or `adviceId`. However, how is that distinct from `Card.uuid` which is being proposed in #496? I was trying to keep `Card.uuid` as the identifier of the individual instance of the advice, and `Card.source.ruleCode` as the identifier of the specific "CDS Magic" in the CDS Service that generated this card. Open to other recommendations on name, but I wish to maintain that distinction.

How is that problem of coding any different from the opacity problem for a code in any other situation? Why would we propose that CDS Hooks have a specific-case solution to what is a generic problem with coding? 

I want to push back against the 'behavioral requirement' suggested. 
* You are making assumptions about the general behavior of the EHR in the first place, and then suggesting constraints on that assumed general behavior. However, at least in our EHR, the assumption about the general behavior itself does not hold true.
* In our implementation, if a new CDS request is made against that service, all the currently displayed cards from that service are dropped, before showing the new set of cards returned. So the behavior requirement proposed is moot to begin with. _(Put another way, there is no session or card stack maintained between subsequent calls to the same service)_
* Additionally, I would say CDS Hooks should not mandate the behavior of the EHR in rendering decision support. It should only standardize the exchange of codified decision support. Different EHR implementations vary widely in how CDS Is presented, and each want to integrate CDS Hooks into their existing decision support frameworks, in the best fashion that fits.

brynrhodes:
I like the idea, and it dovetails nicely with some suggested approaches to tracking the source guidance that are currently using the `link` element and pointing to the underlying guidance. Would it make sense to align with that trajectory and say it should be `guidanceUri` or something like that? With the idea being that it may reference a URL on which the guidance is based (like a reference to a published guideline), or it may be the canonical URL of a knowledge artifact in FHIR, like a PlanDefinition or Library, or it may even be just a web page with more information about the rule?

brynrhodes:
I also agree that it shouldn't make any assumptions about the behavior and duplicate detection/elimination; that's a much broader problem that this could provide input into, but it's only one factor of many.

brynrhodes:
Are we talking about a "category"/"topic"?: http://build.fhir.org/plandefinition-definitions.html#PlanDefinition.topic

mattvarghese:
@brynrhodes - if it is a "coding" then a CDS vendor could use URIs / URLs as code values. So I would prefer we keep it a "coding" with a code-system and a code-value on it rather than a URI. Another system might use GUIDs for all their categories/topics? 

I'm open to calling it topic, as long as it is still agreed between the CDS Service and CDS Client, and not enumerated by the Spec.

docs/specification/1.0.md
```diff
@@ -464,6 +464,23 @@ Field | Optionality | Type | Description
 <nobr>`label`</nobr>| REQUIRED | *string* | A short, human-readable label to display for the source of the information displayed on this card. If a `url` is also specified, this MAY be the text for the hyperlink.
 `url` | OPTIONAL | *URL* | An optional absolute URL to load (via `GET`, in a browser context) when a user clicks on this link to learn more about the organization or data set that provided the information on this card. Note that this URL should not be used to supply a context-specific "drill-down" view of the information on this card. For that, use `link.url` instead.
 `icon` | OPTIONAL | *URL* | An absolute URL to an icon for the source of this card. The icon returned by this URL SHOULD be a 100x100 pixel PNG image without any transparent regions.
+`topic` | OPTIONAL | *[FHIR Coding](http://hl7.org/fhir/datatypes.html#coding)* | A "FHIR Coding" json object containing a descriptive topic related to the content of the card. A topic provides a high-level categorization that can be useful for filtering and searching.
```
dennispatterson: Categorization could clearly be helpful.  Should we say _something_ in here about where these Codings come from, even if it's around mutual agreement between clients/servers.  There will be questions around this field, for sure, but having no comments about coding systems or where this comes from makes it more ambiguous.

docs/specification/1.0.md
```diff
@@ -464,6 +464,23 @@ Field | Optionality | Type | Description
...
+
+Below is an example `source` parameter:
+
+```json
+{
+  "source" : {
```
dennispatterson: Does the new field pertain to where the guidance comes from or what kind of guidance is provided?  Put another way, is this applicable to the source of the card's information or metadata for the card itself?  Couldn't a given source return different variations in decision support 'category'/'topic'?

brettmarquard: I don't think this should be added until we have a preferred value set.

docs/specification/1.0.md
isaacvetter: Moved conversation to zulip.

mattvarghese:
> `label` | OPTIONAL | *string* | A short, human-readable label to display for this **Topic**.
Can we change "label" to "display" so our custom version of CodeSystem can be in line with FHIR Spec? Also, why are we not comfortable referencing FHIR coding directly, when we reference FHIR resources directly?

docs/specification/1.0.md
isaacvetter: Created [jira](https://jira.hl7.org/browse/FHIR-25746) against Clinical Reasoning to create additional example codesystem and codes for DefinitionTopic.

jmandel:
If we're changing `label` --> `display` here, we should do the same in #513. Both data structures basically represent a FHIR Coding, and I don't care much what we call the text bit, but we should be consistent :)

docs/specification/current.md
```diff
@@ -464,6 +464,35 @@ Field | Optionality | Type | Description
...
+
+
+Below is an example `source` parameter:
+
+```json
+{
+  "source" : {
+    "label" : "Zika Virus Management",
+    "url" : "https://example.com/cdc-zika-virus-mgmt",
+    "icon" : "https://example.com/https://example.com/cdc-zika-virus-mgmt/100.png",
```
jmandel:
```suggestion
    "icon" : "https://example.com/cdc-zika-virus-mgmt/100.png",
```

The example URL is broken

isaacvetter:
@jmandel - I just changed #513 (add override reasons) so that this PR's card.topic.display and that PR's card.overrideReasons.display use the same term.

Issue #502, opened 2019-10-04
Patient-view: encounter access


bvdh:
The patient-view context provide an optional field `encounter `that provides the encounter.id of the current Encounter. 

The current specification states in the section "Prefetch Template" (From <https://cds-hooks.org/specification/1.0/#hook-maturity-model> ):

> Regardless of how the CDS Client satisfies the prefetch templates (if at all), the prefetched data given to the CDS Service **MUST be equivalent to the data the CDS Service would receive if it were making its own call to the CDS Client's FHIR server** using the parameterized prefetch template.

	
So a prefetch query using the id from the `encounter `field returns the Encounter. A query to the FHIR Server will return the same Resource.

Other parts of this specification indicate that it is expected that some information is present in the CDS-Client memory and has not yet been persisted to the FHIR-Server, see section "FHIR resources in context" (https://cds-hooks.org/specification/1.0/#hook-maturity-model):
> Often, context is populated with in-progress or in-memory data that may not yet be available from the FHIR server. For example, imagine a hook, medication-order that is invoked when a user selects a medication durating an order workflow.

The results in the question: wow does one retrieve the current Encounter, if stored in memory?
Does this mean that a CDS Server must make sure the Encounter resource (and other data in-memory) is accessible from the FHIR-server?

isaacvetter:
Hey @bvdh ,

This distinction in the spec is between `context`, which can contain "in-memory" data that may not be available via the FHIR server, and `prefetch`'d data which MUST be the same result if the CDS server executed the prefetch template query against the restful fhir server. 

Generally, it is draft resources supplied via `context` may not be available on the FHIR server. There are no draft resources in `patient-view`'s `context` (unlike `order-sign` and `order-select`). I'd expect that all of the fields in the patient-view hook's context are available on the FHIR server and therefore the currently documented patient-view hook is clearly consistent with the base spec as published. 

What this really means is that if a prefetch template references a FHIR resource that is only "in memory", the CDS client shouldn't honor it. 

@bvdh - what do you think?

brynrhodes:
Add to the patient-view documentation a note that the encounter referenced by the encounterId context field is expected to be available via the FHIR Server (if an endpoint is used) or via prefetch if no endpoint is provided. However, the specification does not explicitly require this behavior and implementations should be aware that integrations may not fully support this expectation.

isaacvetter:
Per wg call conversation:

[...]

Issue #503, opened 2019-10-04
Patient-view and risks for inappropriate advice


bvdh:
Some key CDS-Hooks use cases include:
1. Show card based on FHIR data
2. Show multiple cards relate to the same FHIR/context data, possibly from different CDS-Services

This makes the following scenario's possible:
- When the CDS-Client decides to update information a CDS Sever depends upon, an update to the resource(s) will be present in the scratch-patch of the CDS-Client/EMR.  Other cards are based on the old data and and no longer valid, they might even providing suggestions that can be potentially harm-full to the patient. 	
- A Card is presented that provides a suggestion that solves an issue. The Suggestion is followed. Later, another change overrides that suggestion causing the original issue to re-appear. The current specification does not state the CDS-Service is recalled. It can also not access the information in the CDS-Client context. As a result, the practitioner cannot rely on the CDS-Service to provide guidance.

These scenario's both introduce significant and potentially dangerous situations. This raises the question of what use of the patient-view hook is justified that does not introduce these risks? 
If none, we should not publish the hook. 
If some, we should mention those, mention the risks and state that the patient-view SHALL not be used for such advice.

In practice this severely limits the use the patient-view.  Addressing these issues might very well requiring new testing. As a result, one could even wonder whether the required changes are compatible with a maturity-level of 4.....

isaacvetter:
Hey @bvdh,

At root - you're struggling to define how different CDS clients implement workflow and UI as it relates to the display of external CDS. 

Should CDS Hooks really attempt to define this behavior? Our approach thus far has been to not prescribe cds client UI. 

Would non-normative track recommendations for cds client behavior address your concerns?

Isaac

isaacvetter:
Bas,

Do note that we do try to call out a related risk already in the spec: 

>The risk that a CDS Service could return a decision based on outdated patient data, resulting in a safety risk to the patient.

https://cds-hooks.org/specification/1.0/#security-and-safety

brynrhodes:
Add documentation to note: For more information on patient safety issues see [Security & Safety](https://cds-hooks.hl7.org/1.0/#security-and-safety).

PR #506, opened 2019-10-09
availability of data and safety

isaacvetter:
Intended to address -- 
Fixes: #502, #503

### DIFF
--- a/docs/hooks/patient-view.md
+++ b/docs/hooks/patient-view.md
@@ -22,7 +22,7 @@ Field | Optionality | Prefetch Token | Type | Description
 `patientId` | REQUIRED | Yes | *string* | The FHIR `Patient.id` of the current patient in context
 `encounterId` | OPTIONAL | Yes | *string* | The FHIR `Encounter.id` of the current encounter in context
 
-For information on patient safety issues see [Security & Safety](../../specification/1.0/#security-and-safety).
+The `patient-view` hook allows a CDS Service to provide guidance based on information acquired using a prefetch template and the FHIR server. A variety of factors could cause this information to become outdated. Stale clinical data could result in a risk to patient safety. For information on patient safety issues see [Security & Safety](../../specification/1.0/#security-and-safety).
 
 ### Examples
 
--- a/docs/hooks/patient-view.md
+++ b/docs/hooks/patient-view.md
@@ -14,7 +14,7 @@ The user has just opened a patient's record.
 
 The patient whose record was opened, including their encounter, if applicable. 
 
-The FHIR ids referenced in `context` are expected to be available via the FHIR Server (if `fhirServer` is populated in the request) or via prefetch, if supported. However, the specification does not explicitly require this behavior and implementations should be aware that integrations may not fully support this expectation.
+The FHIR resources, referenced by ids in `context`, SHOULD be available to the CDS service. If supported by the CDS Client, the CDS service may request these resources through prefetch. Alternatively or additionally the CDS Client may provide a `fhirserver` in the request for RESTful access.  Support for both `prefetch` and `fhirServer` are optional. As such, CDS Service implementers need to be aware that CDS Clients may not support this behavior.
 
 Field | Optionality | Prefetch Token | Type | Description
 ----- | -------- | ---- | ---- | ----
--- a/docs/hooks/patient-view.md
+++ b/docs/hooks/patient-view.md
@@ -12,13 +12,17 @@ The user has just opened a patient's record.
 
 ## Context
 
-The patient whose record was opened, including their encounter, if applicable.
+The patient whose record was opened, including their encounter, if applicable. 
+
+The FHIR ids referenced in `context` are expected to be available via the FHIR Server (if `fhirServer` is populated in the request) or via prefetch, if supported. However, the specification does not explicitly require this behavior and implementations should be aware that integrations may not fully support this expectation.
 
 Field | Optionality | Prefetch Token | Type | Description
 ----- | -------- | ---- | ---- | ----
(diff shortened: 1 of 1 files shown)


docs/hooks/patient-view.md
```diff
@@ -12,14 +12,18 @@ The user has just opened a patient's record.
 
 ## Context
 
-The patient whose record was opened, including their encounter, if applicable.
+The patient whose record was opened, including their encounter, if applicable. 
+
+The FHIR ids referenced in `context` are expected to be available via the FHIR Server (if `fhirServer` is populated in the request) or via prefetch, if supported. However, the specification does not explicitly require this behavior and implementations should be aware that integrations may not fully support this expectation.
```
bvdh: Suggest:
If the CDS-Client supports pretech and/or provides a FHIR Server, the CDS-Client SHOULD provide FHIR Resources referenced by FHIR ids in `context`. As this behavior is optional, CDS-Server implementers need to be aware that not all CDS-Clients will support this behavior.

docs/hooks/patient-view.md
```diff
@@ -12,14 +12,18 @@ The user has just opened a patient's record.
...
 
 Field | Optionality | Prefetch Token | Type | Description
 ----- | -------- | ---- | ---- | ----
 `userId` | REQUIRED | Yes | *string* | The id of the current user.<br />For example, if the user represents a FHIR resource on the given FHIR server, the resource type would be one of [Practitioner](https://www.hl7.org/fhir/practitioner.html), [Patient](https://www.hl7.org/fhir/patient.html), or [RelatedPerson](https://www.hl7.org/fhir/relatedperson.html).<br />If the user was a Practitioner, this value would be `Practitioner/123`
 `patientId` | REQUIRED | Yes | *string* | The FHIR `Patient.id` of the current patient in context
 `encounterId` | OPTIONAL | Yes | *string* | The FHIR `Encounter.id` of the current encounter in context
 
+For information on patient safety issues see [Security & Safety](../../specification/1.0/#security-and-safety).
```
bvdh: Suggest:  
The patient-view hook only allows advice that is based on information provided using prefetch and/or the FHIR-server. As is stated in  [Security & Safety], implementers should be aware that using the patient-view hook might result in a decision based on outdated patient data, resulting in a safety risk to the patient. E.g. because the data the advice has been based on has been altered in the current user-session.

docs/hooks/patient-view.md
dennispatterson: Is it appropriate to add a statement to the base spec about the nature of context?  e.g. if ids are provided, they must be resolvable through some mechanism available to the CDS Service.  The point being that if an id is present, in order to be useful, the CDS Service must be able to call the FHIR server to retrieve it, or to have that same data prefetched by the CDS Client.

docs/hooks/patient-view.md
dennispatterson: Does this addition provide any information that is not already available in the base spec?

docs/hooks/patient-view.md
isaacvetter: Hey Bas, does the rewrite address your concerns? I modified your suggestion for readability and consistency in terminology.

docs/hooks/patient-view.md
isaacvetter: Dennis - that's a good question. Bas, how much of the base spec should we be repeating here?

docs/hooks/patient-view.md
dennispatterson: First, a few relevant sections of the spec:

* On [CDS Hooks Anatomy](https://cds-hooks.hl7.org/1.0/#cds-hooks-anatomy), a `hook` is 
  > A defined point within the client system's workflow with well-known contextual information provided as part of the request

* On the [HTTP Request](https://cds-hooks.hl7.org/1.0/#http-request_1) to a CDS Service, `context` is 
  > Hook-specific contextual data that the CDS service will need.

* On [Providing FHIR Resources to a CDS Service](https://cds-hooks.hl7.org/1.0/#providing-fhir-resources-to-a-cds-service)
  > Each CDS Service will require specific FHIR resources in order to compute the recommendations the CDS Client requests...this specification defines a process to allow a CDS Service to request and obtain FHIR resources efficiently.  Two optional methods are provided....Each CDS Client SHOULD decide which approach, or combination, is preferred.

* On [Prefetch Template](https://cds-hooks.hl7.org/1.0/#prefetch-template)
  > If the CDS Service is unable to obtain required data because it cannot access the FHIR server and the request did not contain the necessary prefetch keys, the service SHALL respond with an HTTP 412 Precondition Failed status code.

In summary, the spec emphasizes that:
1. A CDS Client's purpose in invoking a hook is to communicate well-known and needed information to the CDS Service
2. CDS Clients SHOULD decide whether to support exposing their FHIR server or implementing prefetch or both.
3. CDS Services SHALL return a 412 if they do not get the information that they need from either the FHIR Server or prefetch (as opposed to returning something unsafe like an empty array of cards).

It isn't the responsibility of a _hook definition_ to warn CDS Services about CDS Clients that choose to limit decision support by preventing access to needed data (or otherwise). Such CDS Clients will not gain adoption and I don't think statements on a hook definition to address their unusable nature add much value.

If we want to add clarifications to a future version of the spec to comment on the nature of FHIR references in a given `context`, that or other similar updates may be helpful.  In the meantime, I don't think the 1.0 specification precludes the ability to publish the existing patient-view definition (prior to these amendments).

@bvdh, thoughts?

docs/hooks/patient-view.md
dennispatterson: Adding comments about outdated/stale clinical data and a link to the Security and Safety page is redundant since the 1.0 specification [already calls this out](https://cds-hooks.hl7.org/1.0/#security-and-safety)

> Security and safety risks associated with the CDS Hooks API include:
> ...
> 6. The risk that a CDS Service could return a decision based on outdated patient data, resulting in a safety risk to the patient.

@bvdh, are you agreeable to removing this language from the hook since it's already covered by the published base spec?

isaacvetter:
Hey Guys - @bvdh , @dennispatterson , 

I'm closing this PR in favor of the HL7 patient-view hook ballot.

Isaac

Issue #500, opened 2019-10-02
transition from 1.0 to current


isaacvetter:
* update menu from 1.0 to current
* Tag commit [baf0339a85f624e1ee141bf119f7051265bd6fc5](https://github.com/cds-hooks/docs/commit/baf0339a85f624e1ee141bf119f7051265bd6fc5#diff-7971a89e057af84349a92f9ee9064c66) as 1.0 release in github.
* Do we need to update the banner on spec page?

bvdh:
I think we should.

Guidance can be found in Guidance-to-work-groups-on-DSTU-updates ([http://gforge.hl7.org/gf/download/docmanfileversion/7482/10824/Guidance-to-Work-Groups-on-DSTU-Updates20130819.docx](url))
As this is a change to an STU that is not balloted, the version should be updated with a subversion. In this case 1.1
 
There is also guidance on the format of the specification, the following text from [http://www.hl7.org/documentcenter/public_temp_E51796A3-1C23-BA17-0CAF66809F735B1E/membership/HL7_Governance_and_Operations_Manual.pdf](url) provides some guideance on this: 
 
**13.02.07.01 Standard for Trial Use (STU) Front Matter** 
The front cover of a STU shall include the following statement. “Publication of this standard for trial use (STU) has been approved by Health Level Seven International (HL7). This STU is not an accredited American National Standard. The comment period for use of this STU shall end [indicate the number of months allocated for evaluation and review] months from the date of publication. Suggestions for revision should be submitted at http://www.hl7.org/stucomments/index.cfm . HL7 Governance and Operations Manual Adopted March 7, 2019 49 Last Update: August 28, 2019 Following this [the number of months shown above] month evaluation period, this STU, revised as necessary, will be submitted to a normative ballot in preparation for approval by ANSI as an American National Standard. Implementations of this STU shall be viable throughout the normative ballot process and for up to six months after publication of the relevant normative standard.” The use of the ANSI logo or trademark on a STU is strictly prohibited and at no time shall the STU be referred to as an American National Standard for Trial Use.
 
So this text must be added to the front page of the spec. As this is STU, the version should be prefixed with STU.
 
So the version of the new specification must  be STU1.1.

brynrhodes:
Rather than calling it 1.1, we should rename it to current, since it is the current draft version of the specification.

isaacvetter:
This issue has been addressed and is no longer relevant.

PR #511, opened 2019-11-18
#500: Renamed 1.0 to current and added 1.0 link to the published spec…

brynrhodes:
…ification on HL7.org

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -14,7 +14,7 @@ nav:
 - Overview: 'index.md'
 - Specification:
     - Current (draft): 'specification/current.md'
-    - '1.0 (External Link)': https://cds-hooks.hl7.org/1.0/
+    - '1.0 (external link)': https://cds-hooks.hl7.org/1.0/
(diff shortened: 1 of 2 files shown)


brynrhodes:
Agreed, added draft label in the menu and on the page for consistency

mkdocs.yml
```diff
@@ -13,7 +13,8 @@ extra:
 nav:
 - Overview: 'index.md'
 - Specification:
-    - '1.0': 'specification/1.0.md'
+    - Current (draft): 'specification/current.md'
+    - '1.0 (External Link)': https://cds-hooks.hl7.org/1.0/
```
dennispatterson: Rather than link to 1.0 (and having to add links for each subsequent version), thoughts on linking to the published versions directory
- HL7 Releases (https://cds-hooks.hl7.org/)

isaacvetter: Great point! Dennis.

Let's do exactly this, once we have more than one release.

PR #513, opened 2019-11-26
overrideReasons for 1.1

isaacvetter:
Add ackReason from https://github.com/cds-hooks/docs/wiki/Acknowledgement-reasons-for-1.1


docs/specification/1.0.md
```diff
@@ -523,6 +524,33 @@ The following example illustrates a delete action:
...
 
+
+
+#### ackReason
+
+An **ackReason** is described by the following attributes.
+
+Field | Optionality | Type | Description
+----- | ----- | ----- | --------
+`key` | REQUIRED | *string* | An opaque, unique identifier for this acknowledgement reason. `key` MUST be unique for the card.  To enable analysis of a card's usage, the same key accompanies the same label across CDS Hooks requests for the same CDS client.
```
```suggestion
`key` | REQUIRED | *string* | An opaque, unique identifier for this acknowledgement reason. `key` MUST be unique within the card.  To enable analysis of a card's usage, the same key accompanies the same label across CDS Hooks requests for the same CDS client.
```

jmandel:
I've read through the text here but I can't figure out what's going on. An end-to-end example might help.

Question I have, that I hope this PR can address:

* Is "acknowledgement" *purely* about indicating why a card's advice was not taken? Or is this just one use case? ("MAY" is confusing me, and the general English-language concept seems to imply positive and negative acknowledgements.)

* Is "ack" about rejecting / ignoring a whole card? It's not clear to me that card-level is the right idea here, but an example with multiple suggestions + some info would probably help.

* When/how is an acknowledgement sent from the client to the service? Is this part of an analytics mechanism?

docs/specification/1.0.md
```diff
@@ -452,6 +452,7 @@ Field | Optionality | Type | Description
...
 `source` | REQUIRED | *object* | Grouping structure for the **Source** of the information displayed on this card. The source should be the primary source of guidance for the decision support the card represents.
 <nobr>`suggestions`</nobr> | OPTIONAL | *array* of **Suggestions** | Allows a service to suggest a set of changes in the context of the current activity (e.g.  changing the dose of the medication currently being prescribed, for the `medication-prescribe` activity). If suggestions are present, `selectionBehavior` MUST also be provided.
+`ackReasons` | OPTIONAL | *array* of **ackReasons** | The CDS service MAY return a list of acknowledgement reasons to the CDS client. The CDS client MAY present these reasons to the clinician when they dismiss a card, enabling the clinician to indicate why the guidance was not taken. A CDS client MAY augment the acknowledgement reasons presented to the user with its own reasons.
```
isaacvetter:
```suggestion
`ackReasons` | OPTIONAL | *array* of **ackReasons** | Acknowledgement reasons can be used by the end user when overriding a card without taking the suggested follow-ups. The CDS service MAY return a list of acknowledgement reasons to the CDS client. The CDS client MAY present these reasons to the clinician when they dismiss a card. A CDS client MAY augment the acknowledgement reasons presented to the user with its own reasons.
```

isaacvetter:
Josh,

Your lack of understanding reflects poorly on my spec authoring. :( Particularly, if you've read the recently created [Acknowledgement-reasons-for-1.1]( https://github.com/cds-hooks/docs/wiki/Acknowledgement-reasons-for-1.1) wiki page.  (Have you read this page?)

Regardless, the spec needs to clearly set expectation for this feature. 

>Is "acknowledgement" purely about indicating why a card's advice was not taken? Or is this just one use case? ("MAY" is confusing me, and the general English-language concept seems to imply positive and negative acknowledgements.)

Yes, an "acknowledgement" is only used to indicate why a card's advice was not taken. I was initially calling these _reject reasons_, we got feedback during the CDS Hooks breakout that a clinician would rather _acknowledge_ cds than reject it. Also, acknowledge is an existing term and reject would be a new term. 

I added a new sentence to the ackReason row within the Card table to try to better explain this. 

>Is "ack" about rejecting / ignoring a whole card? It's not clear to me that card-level is the right idea here, but an example with multiple suggestions + some info would probably help.

Yes, an acknowledgement rejects/ignores the whole card. I think that the card is the right level. An acknowledgement reason is almost an alternate suggestion. The user can pick suggestion A, or suggestion B or the ackReason. 

I struggled to determine how much of the [wiki page]( https://github.com/cds-hooks/docs/wiki/Acknowledgement-reasons-for-1.1) could be forced into the spec. **Are you suggesting that an example should be in the spec?**

>When/how is an acknowledgement sent from the client to the service? Is this part of an analytics mechanism?

Yes, I haven't written the PR content for the "feedback service" yet. Here's the draft wiki writeup: [Feedback-endpoint-for-CDS-Hooks-1.1](https://github.com/cds-hooks/docs/wiki/Feedback-endpoint-for-CDS-Hooks-1.1).

Isaac

isaacvetter:
Following conversation with @dennispatterson , @brettmarquard, we should:
1) Change name of ackReason to overrideReason for simplicity in language and readability. 
2) Consider if a standardized or example set of override reasons is possible or reasonable for the spec to suggest. We think this isn't reasonable today, but should be something to work towards int he future. 
3) Dennis: Change `'key` to `id`. Brett thinks this should be a FHIR code, with an example set. (Are there overrides in clinical reasoning?)

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
 }
 ```
 
+
+
+#### overrideReason
```
brynrhodes: Suggest this should be capitalized `OverrideReason` since it is the name of the type

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+
+An **overrideReason** is described by the following attributes. No standard set of **overridereasons** exists today; implementers are encouraged to submit suggestions for standardization. To enable analysis of a card's usage, the CDS service SHALL send the same `code`, `codesystem`, and `label` across CDS Hooks requests for the same CDS client.
```
brynrhodes: Capitalize the first reference, but remove the formatting from the second:
An **OverrideReason** is described by the following attributes. No standard set of override reasons exists today;

brynrhodes: Instead of "No standard set of override reasons exists today", maybe "This specification does not prescribe a standard set of override reasons;

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+
+Field | Optionality | Type | Description
+----- | ----- | ----- | --------
+`code` | REQUIRED | *string* | A code for this **overridereason**. `code` MUST be unique within the card.  To enable analysis of a card's usage, the CDS service SHALL send the same `code` accompanied by the same `label` across CDS Hooks requests for the same CDS client.
```
brynrhodes: Capitalize **OverrideReason**

brynrhodes: Account for system in this text: "SHALL send the same `code` and `system` accompanied by the same `label` across CDS..."

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+`system` | OPTIONAL | *string* | A codesystem for this **overridereason** `code`.
```
brynrhodes: Capitalize **OverrideReason**

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+`label` | REQUIRED | *string* | A short, human-readable label to display for this override (e.g. the CDS Client might render this as the text on a button).
+
+
+```json
+{
+   "ackReasons":[
```
brynrhodes: Should be "overrideReasons"

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+      {
+         "code":"reason-id-provided-by-service",
+	 "system":"urn:oid:1.2.840.46474.1.3.27.11.6.4.373483.0",
```
brynrhodes: Should use a code system URL here, not an OID... suggest a made up one: `http://example.org/cds-services/fhir/CodeSystem/override-reasons`

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+         "label":"Patient refused"
+      },
+      {
+         "code":"12354",
+	 "system":"urn:oid:1.2.840.46474.1.3.27.11.6.4.373483.0",
```
brynrhodes: Same comment for system

docs/specification/1.0.md
isaacvetter: good catch!

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
 }
 ```
 
+
+
+#### OverrideReason
+
+An **OverrideReason** is described by the following attributes. This specification does not prescribe a standard set of override reasons; implementers are encouraged to submit suggestions for standardization. To enable analysis of a card's usage, the CDS service SHALL send the same `code`, `codesystem`, and `label` across CDS Hooks requests for the same CDS client.
```
dennispatterson: "codeSystem" should be "system"

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
...
+
+Field | Optionality | Type | Description
+----- | ----- | ----- | --------
+`code` | REQUIRED | *string* | A code for this **Overridereason**. `code` MUST be unique within the card.  To enable analysis of a card's usage, the CDS service SHALL send the same `code` and 'system' accompanied by the same `label` across CDS Hooks requests for the same CDS client.
```
dennispatterson: nit: change quotes around 'system' so it's rendered as `system`

docs/specification/1.0.md
```diff
@@ -452,6 +452,7 @@ Field | Optionality | Type | Description
...
 `source` | REQUIRED | *object* | Grouping structure for the **Source** of the information displayed on this card. The source should be the primary source of guidance for the decision support the card represents.
 <nobr>`suggestions`</nobr> | OPTIONAL | *array* of **Suggestions** | Allows a service to suggest a set of changes in the context of the current activity (e.g.  changing the dose of the medication currently being prescribed, for the `medication-prescribe` activity). If suggestions are present, `selectionBehavior` MUST also be provided.
+`overrideReasons` | OPTIONAL | *array* of **OverrideReason** | Override reasons can be selected by the end user when overriding a card without taking the suggested follow-ups. The CDS service MAY return a list of override reasons to the CDS client. The CDS client SHOULD present these reasons to the clinician when they dismiss a card. A CDS client MAY augment the override reasons presented to the user with its own reasons.
```
dennispatterson: nit: move this below selectionBehavior.  This keeps suggestions and behavior close, followed by what to do when a suggestion is not selected.

docs/specification/1.0.md
```diff
@@ -523,6 +524,36 @@ The following example illustrates a delete action:
 }
 ```
 
+
+
+#### OverrideReason
+
+An **OverrideReason** is described by the following attributes. This specification does not prescribe a standard set of override reasons; implementers are encouraged to submit suggestions for standardization. To enable analysis of a card's usage, the CDS service SHALL send the same `code`, `system`, and `label` across CDS Hooks requests for the same CDS client.
```
dennispatterson: "...across CDS Hooks **responses** for the same..."

> the CDS service SHALL send the same `code`, `system`, and `label` across CDS Hooks requests for the same CDS client.

I want to make sure this doesn't get read as each card must have the same code(s), but that the codes/systems/labels are from a _set_ of codes/systems/labels that is consistently drawn from across replies.  What about:

The CDS Service SHALL use a consistent set of override reasons across CDS Hooks responses for the same CDS Client

jmandel: I agree with the confusion; your proposed language here isn't helping me, though. I think we're trying to say: 

1. for a given CDS Client and label the CDS Service SHALL send the same `code` and `system`
2. for a given CDS Client, `code`, and `system`, the CDS Service SHALL send the same `label`

Is that right? i.e., both (1) and (2) are desired? Why is this important? What is the connection to analytics?

docs/specification/current.md
```diff
@@ -453,6 +453,7 @@ Field | Optionality | Type | Description
...
 <nobr>`suggestions`</nobr> | OPTIONAL | *array* of **Suggestions** | Allows a service to suggest a set of changes in the context of the current activity (e.g.  changing the dose of the medication currently being prescribed, for the `medication-prescribe` activity). If suggestions are present, `selectionBehavior` MUST also be provided.
 `selectionBehavior` | OPTIONAL | *string* | Describes the intended selection behavior of the suggestions in the card. Allowed values are: `at-most-one`, indicating that the user may choose none or at most one of the suggestions;`any`, indicating that the end user may choose any number of suggestions including none of them and all of them. CDS Clients that do not understand the value MUST treat the card as an error.
+`overrideReasons` | OPTIONAL | *array* of **OverrideReason** | Override reasons can be selected by the end user when overriding a card without taking the suggested follow-ups. The CDS service MAY return a list of override reasons to the CDS client. The CDS client SHOULD present these reasons to the clinician when they dismiss a card. A CDS client MAY augment the override reasons presented to the user with its own reasons.
```
dennispatterson: Change 'suggested follow-ups' to 'clinical guidance'.

dennispatterson:
I recently had some e-mail conversations with Anna Dover at First Databank to get CDS Service perspective on this, which some comments bleeding into the territory of the [feedback endpoint](https://github.com/cds-hooks/docs/wiki/Feedback-endpoint-for-CDS-Hooks-1.1).  I wanted to share what has been discussed so far:

> Yes, we would return a set of override reasons if possible.  We would very much like to have the ping back of how a user responded to a hook...it might be good to know what action(s) was taken as part of the response.

> The key is the display to the end-user to help illicit the appropriate action.  If they are too specific, then we will have many.  If they are less specific, we will need the ability to overwrite the display but store the ping back with the approved reason.

As for the types of reasons, she shares some but also draws a distinction that not all reasons are _override reasons_, per se.

> I think it could vary a bit from card to card and some vendors will want to have many reasons to help better gain insight in the workflow.  The challenge with calling it an override is that they may be interacting with the card and making a change that is not directly from the card.  Some examples I would want to have available: Lab Ordered, Will Review Orders, Follow-up with Attending, etc.  Some that are overrides are Not Clinically Relevant, New Information Available, etc.

> An example would be “Will Review Clinical Data”, a reason we have with our current integration.  Very often a clinical end-user will not have all of the information they need to process to make a decision about the information in the card.  This allows them to move past the card and review the patient holistically, which may or may not require the intervention based on risk/benefit to the patient. 

> Override implies it was not a meaningful alert. Perhaps we call it a response vs a reason. It is a way to approximate their response to the alert, even if they cannot take an action directly from the hook.  Very often a clinician may review the patient based on an alert, but not make a change. The alert is meaningful but no change is needed at that time.  Monitoring can continue and the review was warranted, even if a change was not.

docs/specification/current.md
```diff
@@ -524,6 +525,34 @@ The following example illustrates a delete action:
...
+`code` | REQUIRED | *string* | A code for this **OverrideReason**.
+`system` | OPTIONAL | *string* | A codesystem for this **OverrideReason** `code`.
+`label` | REQUIRED | *string* | A short, human-readable label to display for this override (e.g. the CDS Client might render this as the text on a button).
+
+
+```json
+{
+   "overrideReasons":[
+      {
+         "code":"reason-id-provided-by-service",
```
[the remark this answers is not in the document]
isaacvetter: 👍

PR #521, opened 2020-03-17
Update dependencies & python version

isaacvetter:
Update dependencies to latest and greatest. Move from Python 2 to 3.


isaacvetter:
Hey Guys -- just looking for a sanity check to ensure that the new version of python and our dependencies builds on your machine. It works on my machine and (more importantly) TravisCI. 

Isaac

PR #522, opened 2020-05-01
Added additional language to appointment-book hook description as per…

lmckenzi:
… Isaac's request: https://jira.hl7.org/browse/FHIR-22236

### DIFF
--- a/docs/hooks/appointment-book.md
+++ b/docs/hooks/appointment-book.md
@@ -8,7 +8,7 @@
 
 ## Workflow
 
-This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It would be triggered for an appointment with the same user, someone else from the same organization as well as when they book an appointment for a patient with someone outside their organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.  
+This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be triggered for an appointment with the appointment creator, a clinician within the same organization as the appointment creator or even for an appointment outside the creator's organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  This hook enables CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.  
 
 ## Context
 
@@ -238,4 +238,3 @@ Field | Optionality | Prefetch Token | Type | Description
 Version | Description
 ---- | ----
 1.0 | Initial Release
-
--- a/docs/hooks/appointment-book.md
+++ b/docs/hooks/appointment-book.md
@@ -8,7 +8,7 @@
 
 ## Workflow
 
-This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be triggered for an appointment with the appointment creator, a clinician within the same organization as the appointment creator or even for an appointment outside the creator's organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  This hook enables CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.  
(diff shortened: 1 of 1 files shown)


docs/hooks/appointment-book.md
```diff
@@ -8,7 +8,7 @@
 
 ## Workflow
 
-This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
+This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It would be triggered for an appointment with the same user, someone else from the same organization as well as when they book an appointment for a patient with someone outside their organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
```
isaacvetter, line 11:
```suggestion
This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be triggered for an appointment with the appointment creator, a clinician within the same organization as the appointment creator or even for an appointment outside the creator's organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  This hook enables CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.  
```

isaacvetter: Small changes for clarity. 

Lloyd, are you certain we need this part:
>and/or any time between those two points

Clarifying that `patient-view` only triggers once during a user's single session with a patient's chart took a while. By suggesting that `appointment-book` may or may not trigger multiple times during the booking of a single appointment will introduce confusion and hamper successful use between implementers.

lmckenzi: My understanding had been that most hooks would continuously trigger as information was being filled in so that a CDS system could intervene as early in the workflow as possible (i.e. once all data was in place that would allow a decision to send a card was available).  The alternative is to allow the user to progress a considerable way down the pathway of completing the details of their action only to have a decent portion of their work no longer available.  This is the same principle as I believe is followed with the 'order' hook, though in that case there's a split between `order-select` and `order-sign`.  It may be that's what's appropriate to do here too - i.e. one hook that triggers continuously with potentially partial information and one hook that triggers just once when all information is available.

isaacvetter: Sorry, this has moved so slow, Lloyd. There are potential performance concerns with continuously triggering a hook, but ultimately, this behavior will depend upon implementors and you're language leaves that open. I'll badger the rest of the team to review this PR as well.

isaacvetter:
@dennispatterson , @jmandel , @brynrhodes - Can you take a look at this small change to `appointment-book`'s definition?

docs/hooks/appointment-book.md
```diff
@@ -8,7 +8,7 @@
 
 ## Workflow
 
-This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be invoked at the start and end of the booking process and/or any time between those two points.  It allows CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
+This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  It may be triggered for an appointment with the appointment creator, a clinician within the same organization as the appointment creator or even for an appointment outside the creator's organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  This hook enables CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.
```
dennispatterson, line 11: > It may be triggered for an appointment with the appointment creator, a clinician within the same organization as the appointment creator or even for an appointment outside the creator's organization.

Is the intention here to constrain the types of appointments for which this can be triggered or merely to give examples?  If it's the latter, can we make that clearer?  "Examples include an appointment with..." or something along those lines?

isaacvetter, line 11:
```suggestion
This hook is invoked when the user is scheduling one or more future encounters/visits for the patient.  For example, the `appointment-book` hook may be triggered for an appointment with the appointment creator, a clinician within the same organization as the appointment creator or even for an appointment outside the creator's organization.  It may be invoked at the start and end of the booking process and/or any time between those two points.  This hook enables CDS Services to intervene in the decision of when future appointments should be scheduled, where they should be scheduled, what services should be booked, to identify actions that need to occur prior to scheduled appointments, etc.  
```

isaacvetter: @dennispatterson - updated, re-review?

PR #523, opened 2020-05-14
Added some clarifications

euvitudo:
Fixed typo
Moved reference to authentication to main Feedback section
Added a note about what type of payload the feedback endpoint accepts
acceptedSuggestions should be an array of string (reducing redundancy in
an object with an 'id'
Added note that each acceptedSuggestion value uniquely identifies the
suggestion
Added note to 'overrideReason' that it is an optional element for
'overridden' outcomes.
Some header changes (### instead of ##)
updated 'Suggestion accepted' example
Moved 'Card ignored' to the end of the Feedback section, as the proposed
spec does not provide support for reporting ignored cards
Changed 'Overridden guidance' to 'Card overridden', for consistency with
the (assumed) intention of the proposed spec
Updated/added 'override or reject' or 'overridden or rejected', as it's
not clear that each term is synonymous
Changed 'Explicit reject with override reasons' to 'Card override with
reasons' for consistency with the (assumed) intention of the proposed
spec
Updated formatting of 'Card override with reasons' example

### DIFF
--- a/docs/specification/current.md
+++ b/docs/specification/current.md
@@ -656,11 +656,26 @@ Each **Feedback** is described by the following attributes.
 
 Field | Optionality | Type | Description
 ----- | ----- | ----- | --------
-`card` | REQUIRED | *string* | The `card.uuid` from the CDS Hooks response. Uniquely identifies the card.
+`card` | REQUIRED | *object* | A json object identifying the card.
 `outcome` | REQUIRED | *string* | A value of `accepted` or `overridden`.
-`acceptedSuggestions` | CONDITIONAL | *array* | An array of string identifying one or more suggestions accepted by the user. Uniquely identifies each suggestion. Required for `accepted` outcomes.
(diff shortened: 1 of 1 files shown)


euvitudo:
This PR was a result of an attempt at implementing the feedback service.

mattvarghese:
@euvitudo, accpetedSuggestion was deliberately made a JSON object with an `id` property in case we want to include additional information on this object. If we keep it as an array of string but find we need to add more info and need to change to an array of objects, wouldn't that be a breaking change?

euvitudo:
@mattvarghese Agreed that it would be a breaking change.  I wasn't paying attention to the decision regarding `id` so that's on me.  For consistency, it seems you might also want `card` to be an object.

mattvarghese:
@euvitudo Why do you say `card` needs to be an array? The feedback request is an array of feedback entries - each one about one particular card.

euvitudo:
@mattvarghese You're right--it would just need to be an object.  (I shouldn't have edited it without thinking.)

euvitudo:
@mattvarghese updated PR.

PR #530, opened 2020-06-01
Fix nesting level for feedback details

jmandel:


isaacvetter:
Non-functional change. Merging with only two approvers.

PR #531, opened 2020-07-09
Enable implementer re-use by creating simple coding data type & move userComment alongside of overrideReason

mattvarghese:
I've created a new CDSHooksCoding type here.
card.source.topic, and card.overrideReasons have been switched to use this new type.
Also, the CDS Hooks Feedback message's overrideReason property has been updated to contain a coding object and a userComment string to be consistent with the rest of the spec.


mattvarghese:
@isaacvetter  , @jmandel  , @dennispatterson  , @brynrhodes  , @brettmarquard   

Hey folks, can get this in before many people start implementing the Feedback service?

docs/specification/current.md
```diff
@@ -555,16 +546,11 @@ The following example illustrates a delete action:
 }
 ```
 
-#### OverrideReason
-
-An **OverrideReason** is described by the following attributes. This specification does not prescribe a standard set of override reasons; implementers are encouraged to submit suggestions for standardization. 
+#### overrideReasons
```
isaacvetter, line 549: Suggest removing this addition. Throughout the rest of the spec we document each data element as if it were a type. For example, `links` is an array of *Link*. The documentation describes *Link*, not `links`. If we want to add this overrideReasons header, we'd need to also add a links, suggestions and cards header.

docs/specification/current.md
```diff
@@ -555,16 +546,11 @@ The following example illustrates a delete action:
...
 
-Field | Optionality | Type | Description
------ | ----- | ----- | --------
-`code` | REQUIRED | *string* | A code for this **OverrideReason**.
-`system` | OPTIONAL | *string* | A codesystem for this **OverrideReason** `code`.
-`display` | REQUIRED | *string* | A short, human-readable label to display for this override (e.g. the CDS Client might render this as the text on a button).
+**overrideReasons** is an array of **CDSHooksCoding** that captures a codified set of reasons an end user may select from as the rejection reason when rejecting the advice presented in the card. When using the coding object representing a reason, implementations are required to only respect the *code* property. However, they may consume other properties for a better end user experience, such as presenting a human readable text in the *display* property instead of the *code* itself to the end user.
```
isaacvetter, line 551:
```suggestion
**overrideReasons** is an array of **Coding** that captures a codified set of reasons an end user may select from as the rejection reason when rejecting the advice presented in the card. When using the coding object representing a reason, implementations are required to only respect the *code* property. However, they may consume other properties for a better end user experience, such as presenting a human readable text in the *display* property instead of the *code* itself to the end user. 
```

docs/specification/current.md
```diff
@@ -655,7 +641,7 @@ Field | Optionality | Type | Description
 `card` | REQUIRED | *string* | The `card.uuid` from the CDS Hooks response. Uniquely identifies the card.
 `outcome` | REQUIRED | *string* | A value of `accepted` or `overridden`.
 `acceptedSuggestions` | CONDITIONAL | *array* | An array of json objects identifying one or more of the user's **AcceptedSuggestion**s. Required for `accepted` outcomes.
-`overrideReason` | OPTIONAL | *object* | A json object identifying the **OverrideReason** selected by the user.
+`overrideReason` | OPTIONAL | *object* of **OverrideReason** | A json object capturing the override reason as well as any comments entered by the user.
```
isaacvetter, line 644:
```suggestion
`overrideReason` | OPTIONAL | **Coding** | A json object capturing the **OverrideReason** as well as any comments entered by the user.
```

docs/specification/current.md
```diff
@@ -721,8 +707,7 @@ Each **OverrideReason** is described by the following attributes, in the feedbac
 
 Field | Optionality | Type | Description
 ----- | ----- | ----- | --------
-`code` | CONDITIONAL | *string* | The code for this OverrideReason as provided by the CDS Service in the CDS Hooks response. Required if user selected an overrideReason (instead of only leaving a `userComment`.
-`system` | CONDITIONAL | *string* | The codesystem for this OverrideReason code as provided by the CDS Service in the CDS Hooks response. Required if the user selected an overrideReason and the CDS Service supplied this information in the CDS Hooks response.
+`reason` | CONDITIONAL | *object* of **CDSHooksCoding** | The Coding object representing the override reason selected by the end user. Required if user selected an override reason from the list of reasons provided in the Card (instead of only leaving a userComment).
```
isaacvetter, line 710:
```suggestion
`reason` | CONDITIONAL |**Coding** | The Coding object representing the override reason selected by the end user. Required if user selected an override reason from the list of reasons provided in the Card (instead of only leaving a userComment).
```

docs/specification/current.md
```diff
@@ -929,6 +917,20 @@ As another example, an extension defined on the discovery response could look li
 ```
 [OAuth 2.0]: https://oauth.net/2/
 
+## Data Types
+
+This section defines some data types that are used across the specification. These types are generic enough so that they may be used with various properties pertaining to different concepts in the text of the specification.
```
isaacvetter, line 922:
```suggestion
This section defines data types used re-used across the specification. 
```

docs/specification/current.md
```diff
@@ -929,6 +917,20 @@ As another example, an extension defined on the discovery response could look li
...
+
+### CDSHooksCoding
```
isaacvetter, line 924:
```suggestion
### Coding
```

docs/specification/current.md
```diff
@@ -929,6 +917,20 @@ As another example, an extension defined on the discovery response could look li
...
+
+The CDSHooksCoding data type captures the concept of a code. A code is understood only when the given code, code-system, and a optionally a human readable display are available. This coding type is a standalone data type in CDS Hooks modeled after a trimmed down version of the [FHIR Coding data type](http://hl7.org/fhir/datatypes.html#Coding).
```
isaacvetter, line 926:
```suggestion
The **Coding** data type captures the concept of a code. A code is understood only when the given code, code-system, and a optionally a human readable display are available. This coding type is a standalone data type in CDS Hooks modeled after a trimmed down version of the [FHIR Coding data type](http://hl7.org/fhir/datatypes.html#Coding).
```

docs/specification/current.md
mattvarghese: This is not correct. The OverrideReason here is a type which contains a reason of type Coding, and a userComment.

docs/specification/current.md
mattvarghese: You have a repeated word in your suggestion.

mattvarghese: I'm accepting this, and then I'll fix the repetition

docs/specification/current.md
mattvarghese: I re-titled this to **Reasons for rejecting a card** since the discussion in that paragraph needs a title separating it from what is above.

Note that there is a separate **OverridReason** type under Feedback that is still necessary.

docs/specification/current.md
mattvarghese: I updated this to just be **OverrideReason**, without the *object* of prefix.

isaacvetter: Ah, sorry, I got confused. Good catch.

docs/specification/current.md
```diff
@@ -733,8 +720,11 @@ POST {baseUrl}/cds-services/{serviceId}/feedback
          "card":"9368d37b-283f-44a0-93ea-547cebab93ed",
          "outcome":"overridden",
          "overrideReason": { 
-	 	"code":"reason-code-provided-by-service",
-     		"system":"http://example.org/cds-services/fhir/CodeSystem/override-reasons",
+	 	"reason": {
+	 		"code":"reason-code-provided-by-service",
+     			"system":"reason-system-provided-by-service",
+     			"display":"reason-display-provided-by-service",
```
dennispatterson, line 726: nit: there isn't any value I can think of to sending the display to the feedback endpoint.  It's certainly benign and shows the full Coding structure here, but it also touches on the difference between making an API call and sending a FHIR resource.  I think the example should reflect that a CDS service just needs to know the machine-readable data.

docs/specification/current.md
```diff
@@ -929,6 +919,20 @@ As another example, an extension defined on the discovery response could look li
 ```
 [OAuth 2.0]: https://oauth.net/2/
 
+## Data Types
+
+This section defines data types re-used across the specification.
```
dennispatterson, line 924:
```suggestion
CDS Hooks leverages json data types throughout.  This section defines data structures re-used across the specification. 
```

isaacvetter:
Per Bryn's review, here's the FHIR [coding](https://www.hl7.org/fhir/datatypes.html#Coding) data type -- why not also include 
* version
* userSelected

docs/specification/current.md
```diff
@@ -929,6 +919,20 @@ As another example, an extension defined on the discovery response could look li
...
+### Coding
+
+The **Coding** data type captures the concept of a code. A code is understood only when the given code, code-system, and a optionally a human readable display are available. This coding type is a standalone data type in CDS Hooks modeled after a trimmed down version of the [FHIR Coding data type](http://hl7.org/fhir/datatypes.html#Coding).
+
+Field | Optionality | Type | Description
+----- | ----- | ----- | --------
+`code` | REQUIRED | *string* | The code for what is being represented
+`system` | OPTIONAL | *string* | A codesystem for this `code`.
+`display` | OPTIONAL | *string* | A short, human-readable label to display.
+
```
isaacvetter, line 935:
```suggestion
`version` | OPTIONAL | * string* | Version of the system - if relevant.
`userSelected` | OPTIONAL | *string* | If this coding was chosen directly by the user.
--



--



'userSelected` | OPTIONAL | *boolean* |
```

docs/specification/current.md
isaacvetter, line 726:
```suggestion
```

docs/specification/current.md
```diff
@@ -733,8 +720,10 @@ POST {baseUrl}/cds-services/{serviceId}/feedback
          "card":"9368d37b-283f-44a0-93ea-547cebab93ed",
          "outcome":"overridden",
          "overrideReason": { 
-	 	"code":"reason-code-provided-by-service",
-     		"system":"http://example.org/cds-services/fhir/CodeSystem/override-reasons",
+	 	"reason": {
+	 		"code":"reason-code-provided-by-service",
+     			"system":"reason-system-provided-by-service",
```
isaacvetter, line 725:
```suggestion
     			"system":"reason-system-provided-by-service"
```

isaacvetter:
@dennispatterson - will you please re-review?

docs/specification/current.md
```diff
@@ -929,6 +918,21 @@ As another example, an extension defined on the discovery response could look li
...
+
+The **Coding** data type captures the concept of a code. A code is understood only when the given code, code-system, and a optionally a human readable display are available. This coding type is a standalone data type in CDS Hooks modeled after a trimmed down version of the [FHIR Coding data type](http://hl7.org/fhir/datatypes.html#Coding).
+
+Field | Optionality | Type | Description
+----- | ----- | ----- | --------
+`code` | REQUIRED | *string* | The code for what is being represented
+`system` | OPTIONAL | *string* | A codesystem for this `code`.
+`display` | OPTIONAL | *string* | A short, human-readable label to display.
+`version` | OPTIONAL | *string* | Version of the system - if relevant.
+`userSelected` | OPTIONAL | *boolean* | If this coding was chosen directly by the user.
```
dennispatterson, line 935: Topics or OverrideReason options returned from the CDS service will never be marked as userSelected.

An override reason sent to the feedback endpoint so far has been echoing back one of the options returned by the CDS service rather than some translation.  Isn’t `userSelected` typically for something like a CodeableConcept, to distinguish whether a specific Coding was directly chosen, to help when translating among multiple synonymous codes from different systems?

docs/specification/current.md
```diff
@@ -929,6 +918,21 @@ As another example, an extension defined on the discovery response could look li
...
+### Coding
+
+The **Coding** data type captures the concept of a code. A code is understood only when the given code, code-system, and a optionally a human readable display are available. This coding type is a standalone data type in CDS Hooks modeled after a trimmed down version of the [FHIR Coding data type](http://hl7.org/fhir/datatypes.html#Coding).
+
+Field | Optionality | Type | Description
+----- | ----- | ----- | --------
+`code` | REQUIRED | *string* | The code for what is being represented
+`system` | OPTIONAL | *string* | A codesystem for this `code`.
+`display` | OPTIONAL | *string* | A short, human-readable label to display.
+`version` | OPTIONAL | *string* | Version of the system - if relevant.
```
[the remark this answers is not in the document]
isaacvetter: Hey @dennispatterson , are we interested in re-using FHIR's coding data type? If so, that will drive inclusion of both version and userSelected. 

Bryn felt that version would be useful to have version and that alignment with FHIR's datatype was good.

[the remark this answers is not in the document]
dennispatterson: FHIR's Coding is applicable across a wide variety of FHIR use-cases.  If we're defining our own type for a limited number of use-cases (this PR even calls it a "trimmed down version"), and FHIR's own documentation says that well-maintained systems don't need version, do we need to start with the presumption that the code systems yet to be defined won't be well-maintained? Or, just add it if becomes necessary?  Note that if a FHIR Coding was passed, it wouldn't be _incompatible_.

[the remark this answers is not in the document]
isaacvetter: Alright, `version` removed.

docs/specification/current.md
isaacvetter: `userSelected` removed. @dennispatterson - do you want to re-review? We need to merge this today.

PR #536, opened 2020-08-19
Introducing sections for coordination

maxsibilla:
We have added a new section under current as well as examples on the oder-select and order-sign pages to explain how configuration options can be used to support the functionality of coordinating between `order-select` and `order-sign` to prevent alert duplication.


docs/specification/current.md
```diff
@@ -1095,6 +1095,76 @@ As each hook progresses through a process of being defined, tested, implemented,
...
+      }, ...
+
+snipped for brevity
+```
+
+##### Extension
+
+###### Configuration Items
+
+The CDS Service MUST process configuration options sent as within an extension resource with the CDS Hooks request in the field `configuration-items`. There are two standart configuration options:  `cache-for-order-sign-filtering` (only used in `order-select`), `filter-out-repeated-alerts` (only used in `order-sign`). All of these configuration options accept Boolean values and are required. The `cache-for-order-sign-filtering` option is used to cache the medication resource during `order-select` and is used as a reference during `order-sign` if `filter-out-repeated-alerts` is set to TRUE. The `filter-out-repeated-alerts` option is used to hide card results if they were shown in the `order-select` results.
```
dennispatterson, line 1161: The 1.0 spec [states](https://cds-hooks.hl7.org/1.0/#extensions):
> The specification is not prescriptive about support for extensions. However, to support extensions, the specification reserves the name extension and will never define an element with that name

The spec allows for extensions but does not define them. 
 Based on that, any kind of functionality like this would need to be directly supported elements that can be applied across any number of hooks.

docs/specification/current.md
```diff
@@ -1095,6 +1095,76 @@ As each hook progresses through a process of being defined, tested, implemented,
...
+
+#### FHIR Server Request
+
+With coordination, the parse and pre-process event for an `order-select` request is nearly identical to what occurs with an `order-select` request without coordination except that the CDS Service checks to see if `cache-for-order-sign-filtering` was set to TRUE and, if so, it caches information that it will later use to filter duplicate CDS Hooks cards from a future `order-sign` request. Then, during the parse and pre-process phase for `order-sign`, the CDS service checks to see if `filter-out-repeated-alerts` is also set to TRUE. If so, it then checks the cached data for a match on specific resources and information to identify if the `order-sign` card response would be a duplication. If so, repeated alert cards will not be returned.
```
dennispatterson, line 1165: Because the base spec is a general framework for communicating between CDS Clients and Services, we should be careful not to add content that is too hook-specific. True, there may be examples of hooks, but attributes such as "cache-for-order-sign-filtering" are specific to ordering hooks and would not be present in the base spec.

docs/specification/current.md
```diff
@@ -1095,6 +1095,76 @@ As each hook progresses through a process of being defined, tested, implemented,
...
+
+This implementation guide discusses a mechanism for providing the CDS service configuration from CDS Hooks requests. The use of this mechanism is required for the functionality of coordinating between order-select and order-sign. There is no requirement to use additional configuration options at this time.
```
dennispatterson, line 1101: > This implementation guide...

Because the base spec is not an implementation guide, any proposals should be more generically applicable across a variety of hooks and usage of those mechanisms would be defined in an implementation guide, such as PDDI.

rkboyce:
Thanks for your comments. I will first respond to the idea of why we think the client should have a role in deciding if alerts are suppressed because of possible duplication between order-select and order-sign. The primary motivation for configuration options is to support the advanced functionality of coordinating between order-select and order-sign to prevent alert duplication. Although it might seem reasonable to have the service default to this kind of coordination with no input from the EHR client, we think that there might be potential scenarios where this form of duplication might be desirable. For example, a teaching hospital might have resident physicians go through the order selection process and then have an attending physician sign off on the order. In this situation, having the same CDS Hooks cards returned at order-select and order-sign would not be perceived as a duplication to the attending. There might be other scenarios where repeating cards across the hooks makes sense. Thus, it seems best to have the EHR client some active role in the decision of whether to filter out alert responses. Using configuration options sent from the EHR client to the service provides a way for the client to have a say in how apparent duplication is handled.

rkboyce:
In reply to Denis' comments that configuration options such as "cache-for-order-sign-filtering" are specific to ordering hooks and would not be present in the base spec. If I understand, we  specific to the order-select and order-sign hooks? If so, we will be happy to edit our branch to remove suggestions from the base and focus these proposed changes to those hooks (those hooks already have edited sections, we would just have to edit to provide more context). That being said, would it still make sense to have a section in the base spec to discuss ideas and support around coordination and point to those hooks for which there is some solution? I think it is a question that will keep coming up.

dennispatterson:
While Matt was reviewing the merits of the feedback service for server-side logic improvements, I think he's still in agreement that the client plays a role in deciding whether alerts are suppressed.  It seems that the distinction is whether the client is able to make a more dynamic decision to filter based on information it has (hook context, attributes in the card like indicator/topic, other attributes not yet present?), or by the client opting into the service's own logic.

The proposed mechanisms to cache and filter are very binary configurations, and so some of the hesitance may also be because of wanting to ensure more factors are considered.  Matt mentioned factoring in topic and severity.  For what context should filtering take place?  That is, what does the service consider a duplicate alert (considering both the card contents and basis for the alert)?   How long should the caching take place?  What's the balance of adding these knobs to services vs providing the CDS Client with enough data to make some of its own decisions?

PR #541, opened 2020-11-11
J#28623: Clarified hook name guidance in hook template

brynrhodes:


isaacvetter (approved):

docs/hooks/template.md
```diff
@@ -8,7 +8,7 @@
```
dennispatterson, line 8: Won't let me comment here directly on the line, but change the hook maturity link above to 'current' instead of '1.0'

docs/hooks/template.md
```diff
@@ -8,7 +8,7 @@
 
 ## Workflow
 
-<mark>Describe when this hook occurs in a workflow. Hook creators SHOULD include as much detail and clarity as possible to minimize any ambiguity or confusion amongst implementors.</mark>
+<mark>Describe when the `hook-name-expressed-as-noun-verb` hook occurs in a workflow. Hook creators SHOULD include as much detail and clarity as possible to minimize any ambiguity or confusion amongst implementers. The hook name should take the form `noun-verb`, such as `encounter-start`, or `order-select`.</mark>
```
dennispatterson, line 11: Can we link this page back to ../../specification/current/#hook-definition-format ?  I'm fine re-stating the noun-verb requirement, but the template should link to where we've documented expectations for what is included.

docs/hooks/template.md
isaacvetter: Dennis, this is potentially messy because the HL7 publish will presumably be to 1.1, not current; however, it also looks like this template page isn't curently published at hl7.org, soo .... made this change.

docs/hooks/template.md
isaacvetter: done.

dennispatterson (approved):

Issue #133, opened 2018-01-02
Extension specification issues


JustinStauffer:
This is in regard to the concept of extensions discussed in issue 76 (https://github.com/cds-hooks/docs/issues/76). Since that issue is closed, I created a new one to track this.

Currently, the CDS Hooks extension model uses a dynamically named element (https://github.com/cds-hooks/docs/commit/b34534ee30c61e99489f86df7662efde65559663).

An example is:
`
"extension": {
   "example-client-conformance": "http://my.hooks.org/fhir/102/Conformance/patientview"
}
`

This dynamically named element will require customization to off-the-shelf JSON libraries because it uses a dynamically named element (e.g. "example-client-conformance") whose name is not determined at design time. For example, if you have a C# .NET class that represents this extension element, it would look like this:

`
public class extension
{
    public string example-client-conformance { get; set; }
}
`
(This example would not work in .NET because the property name has dashes but let's forget that detail for now.)

Now if someone wants to use a different extension name (such as "myextensionproperty"), application developers will need to add another property to their class called "myextensionproperty";

`
public class extension
{
    public string example-client-conformance { get; set; }
    public string myextensionproperty { get; set; }
}
`

That is obviously not sustainable. 

Yes, there are ways one can program around this and everyone will need to do so to handle elements like prefetch. However, this is another case that requires customization and since this piece of the specification is new, we should fix it before it is finalized so we do not require yet more customization to off-the-shelf APIs which makes it more difficult for client and service developers to support this part of the specification.

Instead of the current approach, can we model this more like a FHIR Extension (http://hl7.org/fhir/extensibility.html)? I don't think we have to get _quite as complicated_ as FHIR's extensibility model but using a key/value pair would be much more preferable than using dynamically named JSON properties.

For example, this is readable/writable using a standard JSON formatter/writer and does not require customizing out-of-the-box / off-the-shelf JSON libraries:

`
"services": [
{
    "extension": [{
        "url": "http://domain/example-client-conformance",
        "value": "http://my.hooks.org/fhir/102/Conformance/patientview"
     },
    {
        "url": "http://domain/example-client-another-extension",
        "value": "another-extension-value"
     }]
}
`

Notice that I simplified this a little from the FHIR Extension type in that I am just using one "value" property instead of "valueInt", "valueDecimal", "valueString", etc.

kpshek:
[...]

Issue #89, opened 2017-09-09
Cds service prefetch key is dynamic and makes message serialization hard


isaacvetter:
Here's our sample discovery pre-fetch example:
```
      "prefetch": {
        "patientToGreet": "Patient/{{Patient.id}}"
      }
```
By having the cds service define the "patientToGreet" key, serialization of a CDS Hooks request message by the EHR is many programming languages doesn't allow the use of standard serialization tools.

JustinStauffer:
I brought this concern up to Isaac while investigating CDS Hooks. Having a dynamically named element in the message content essentially means there can be no schema defining the message structure and therefore, you cannot build a code-representation of the message at compile time. That seems like a very big problem to me.

Perhaps you can work around it by creating custom serializers/formatters but that also looks like a big red flag to me. When the solution to a problem with a spec is "Just write a custom _____", that doesn't smell right.

jmandel:
Are you asking for something more like:

```
"prefetch": [{
  "key": "patientToGreet",
  "value": "Patient/{{Patient.id}}"
}]
```

... the same assessment should apply to `.context` and other places where hook-specific variation may be introduced.

kpshek:
Per discussion at the Sept WGM (Tue Q2), this is something we need to decide if we change prior to the 1.0 release (since it would represent a breaking change).

We want to talk with @JustinStauffer & @isaacvetter more on this issue to understand this in further detail.

jaredrn:
I would agree that prefetch is a little tricky for a statically typed language like C#.  It can be overcome with a little custom serialization logic to bridge the gap between a JSON parsing library and the C# reference implementation, but it's not entirely intuitive.

JustinStauffer:
Yes (to Josh / @jmandel). That would work because the _value_ is dynamic and not the element name.

grahamegrieve:
I think that there's some mus-understanding here. 

"prefetch": [ "patientToGreet", "value": "Patient/{{Patient.id}}" ]

this is easy to serialise natively - if you have a definition for *this specific hook*.

JustinStauffer:
@grahamegrieve - Yes but if you want to create a generic framework to handle many hooks, this would become unscalable.

grahamegrieve:
[...]

PR #554, opened 2021-01-15
Fix delete; clarify extension

mattvarghese:
* Deleting a resource should use a resource with just resourceType and id
* Extension examples updated to always have a reverse uri naming convention


ssmillard (commented):
For extensions, Would it be possible to have a structure such as:	"extension": [
		{
			"uri": "com.example.timestamp",
			"value": "2017-11-27T22:13:25Z"
		},
		{
			"uri": "com.cds-hooks.sandbox.myextension-practitionerspecialty",
			"value": "gastroenterology"
		}]

dennispatterson:
@ssmillard we had a bit of discussion about the extensions topic and I dug up https://github.com/cds-hooks/docs/issues/133 and https://github.com/cds-hooks/docs/issues/89, where there was a decision made to handle it this way (for both extensions and prefetch) vs the uri/value pairs mechanism.

isaacvetter:
Related: http://hl7.org/fhir/us/davinci-crd/2019May/hooks.html#propose-alternate-request

The issue with:
1. Deleting a resource should use a resource with just resourceType and id

Is that the same element name (`resource`) using different data types between delete vs create/update is inelegent depending upon the dev stack. 

An alternative to the proposal would be to introduce a new element name (e.g. resourceLocation) to be used for deletes, which could also enable a transition path, and cds service could even provide both. 

If we make any changes, we should ensure that CRD stays in sync as well.

isaacvetter:
@dennispatterson , @brynrhodes and I talked through this issue. 

Ultimately, it is awkward for developers to treat the same `resource` element as either a resource "location" (aka ProcedureRequest/123) or as a full resource. We believe that delete suggestions are not yet widely implemented. Specific to delete suggestions, from a developer perspective, constructing a shell resource withonly a resource type and id 

```
`resourceId` | CONDITIONAL | *string* | A relative reference to the relevant resource in the suggestion. SHOULD be provided for delete suggestions. 
`resource` | OPTIONAL | *object* | A FHIR resource. When the `type` attribute is `create`, the `resource` attribute SHALL contain a new FHIR resource to be created.  For `update`, this holds the updated resource in its entirety and not just the changed fields. Use of this field to communicate a string of a FHIR id for delete suggestions is DEPRECATED and `resourceId` SHOULD be used instead.
```
~~For `delete`, this SHALL be the id of the resource to remove. In hooks where only one "content" resource is ever relevant, this attribute MAY be omitted for `delete` action types only.~~

Todo:
- @isaacvetter : create jira issue

isaacvetter:
Hey @ssmillard,

>For extensions, Would it be possible to have a structure such as: "extension": [

Yes, definitely. The idea is that everything within the extension element is entirely yours to do with as you want.

docs/specification/current.md
```diff
@@ -503,7 +503,8 @@ Field | Optionality | Type | Description
...
 `description` | REQUIRED | *string* | Human-readable description of the suggested action MAY be presented to the end-user.
-`resource` | OPTIONAL | *object* | Depending upon the `type` attribute, a new resource or the id of a resource. When the `type` attribute is `create`, the `resource` attribute SHALL contain a new FHIR resource to be created. For `delete`, this SHALL be the id of the resource to remove. In hooks where only one "content" resource is ever relevant, this attribute MAY be omitted for `delete` action types only. For `update`, this holds the updated resource in its entirety and not just the changed fields.
+`resource` | CONDITIONAL | *object* | A FHIR resource. When the `type` attribute is `create`, the `resource` attribute SHALL contain a new FHIR resource to be created.  For `update`, this holds the updated resource in its entirety and not just the changed fields. Use of this field to communicate a string of a FHIR id for delete suggestions is DEPRECATED and `resourceId` SHOULD be used instead.
+`resourceId` | CONDITIONAL | *string* | A relative reference to the relevant resource in the suggestion. SHOULD be provided for when the `type` attribute is `delete`.
```
dennispatterson, line 507: Minor wordsmithing
```suggestion
`resourceId` | CONDITIONAL | *string* | A relative reference to the relevant resource. SHOULD be provided when the `type` attribute is `delete`. 
```

docs/specification/current.md
```diff
@@ -539,7 +540,8 @@ The following example illustrates a delete action:
 {
 	"type": "delete",
 	"description": "Remove the inappropriate order",
-	"resource": "ProcedureRequest/procedure-request-1"
+	"resourceId": "ServiceRequest/procedure-request-1"
+	}
```
dennispatterson, line 544: Get rid of the extra bracket
```suggestion
```

dennispatterson (changes requested):

mattvarghese:
Hey Folks - are we all agreed on this change? Can we move ahead with this pull request?

dennispatterson (approved):

isaacvetter (approved):

PR #571, opened 2021-08-12
clarify prefetch query search limitation recommendations

isaacvetter:
Fixes FHIR-28634
Putting in own PR, because, I'm not 100% sure that this is the right change.

### DIFF
--- a/docs/specification/current.md
+++ b/docs/specification/current.md
@@ -268,9 +268,8 @@ No single FHIR resource represents a user, rather Practitioner and PractitionerR
 
 To reduce the implementation burden on CDS Clients that support CDS Services, this specification RECOMMENDS that prefetch queries only use a subset of the full functionality available in the FHIR specification. Valid prefetch templates SHOULD only make use of:
 
-* _instance_ level [read](https://www.hl7.org/fhir/http.html#read) interactions (for resources with known ids such as `Patient` and `Practitioner`)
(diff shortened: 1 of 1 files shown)


docs/specification/current.md
```diff
@@ -268,9 +268,8 @@ No single FHIR resource represents a user, rather Practitioner and PractitionerR
 
 To reduce the implementation burden on CDS Clients that support CDS Services, this specification RECOMMENDS that prefetch queries only use a subset of the full functionality available in the FHIR specification. Valid prefetch templates SHOULD only make use of:
 
-* _instance_ level [read](https://www.hl7.org/fhir/http.html#read) interactions (for resources with known ids such as `Patient` and `Practitioner`)
+* _instance_ level [read](https://www.hl7.org/fhir/http.html#read) interactions (for resources with known ids such as `Patient`, `Practitioner`, or `Encounter`; e.g. `patient={{context.patientId}}`)
```
dennispatterson, line 271: "such as" implies an example, so I don't think we need to keep growing the list.  This line is also focused on instance-level retrieval (by id), so it wouldn't involve a search at the type level.

docs/specification/current.md
```diff
@@ -268,9 +268,8 @@ No single FHIR resource represents a user, rather Practitioner and PractitionerR
...
 * _type_ level [search](https://www.hl7.org/fhir/http.html#search) interactions
-* Patient references (e.g. `patient={{context.patientId}}`)
```
dennispatterson, line 273: Do we know the history on this line?  I'm trying to understand what kind of constraints it's meant to communicate. No other resource references in a type search?

PR #572, opened 2021-08-18
fixes: FHIR-28728

isaacvetter:
clarify no-guidance and json nulls
Non-functional change.

### DIFF
--- a/docs/specification/current.md
+++ b/docs/specification/current.md
@@ -7,7 +7,7 @@
 
 The CDS Hooks specification describes the RESTful APIs and interactions to integrate Clinical Decision Support (CDS) between CDS Clients (typically Electronic Health Record Systems (EHRs) or other health information systems) and CDS Services. All data exchanged through the RESTful APIs MUST be sent and received as [JSON](https://tools.ietf.org/html/rfc8259) (JavaScript Object Notation) structures, and MUST be transmitted over channels secured using the Hypertext Transfer Protocol (HTTP) over Transport Layer Security (TLS), also known as HTTPS and defined in [RFC2818](https://tools.ietf.org/html/rfc2818).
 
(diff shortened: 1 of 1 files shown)


docs/specification/current.md
```diff
@@ -7,7 +7,7 @@
...
 
-Unless otherwise specified, JSON attributes SHALL NOT be null. If a JSON attribute is defined with an optionality of OPTIONAL, but does not have a value, implementers MUST omit it. For instance, OPTIONAL JSON string and array attributes should be omitted rather than having a null or empty value. Similarly, JSON objects SHALL NOT be empty.
+Unless otherwise specified, JSON attributes SHALL NOT be null. If a JSON attribute is defined with an optionality of OPTIONAL, but does not have a value, implementers MUST omit it. For instance, OPTIONAL JSON string and array attributes should be omitted rather than having a null or empty value. Similarly, JSON objects SHALL NOT be empty. (Note that CDS Services can return [no guidance with an empty array of cards](#http-response)).
```
dennispatterson, line 10: Instead of re-listing or linking to exceptions, can we just update the first sentence to more clearly apply to the entire paragraph? Something like:
```suggestion
Unless otherwise specified, the following restrictions on null/empty attributes hold true: JSON attributes SHALL NOT be null. If a JSON attribute is defined with an optionality of OPTIONAL, but does not have a value, implementers MUST omit it. For instance, OPTIONAL JSON string and array attributes should be omitted rather than having a null or empty value. Similarly, JSON objects SHALL NOT be empty.
```

It could even be a bulleted list if that makes the coverage clearer.

For the specific example of cards array, it may also be helpful to update the 'cards' attribute description to state that it SHALL be empty if there are no cards, rather than that only be covered in the subsequent example.

PR #599, opened 2021-12-22
propose `order-discontinue`

isaacvetter:
propose `order-discontinue`

### DIFF
--- a/docs/hooks/order-discontinue.md
+++ b/docs/hooks/order-discontinue.md
@@ -1,0 +1,46 @@
+`order-discontinue`
+
+!!! info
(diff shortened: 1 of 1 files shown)


isaacvetter:
John Silva pointed out that the existing definitions differentiate between discontinue and cancel, like so:

>From table 0119 - Order Status codes ( https://www.hl7.org/fhir/v2/0119/index.html )
>"The differentiation between discontinue and cancel is that discontinue effects the order/service and all future occurrences, cancel refers to just the present action."

It seems like both cancel and discontinue are equivalent when the order has only a single occurrence (incl. remaining). Would the request resource sent in `toBeDiscontinued` enable the service to differentiate between these two scenarios? (e.g. send child request for cancels, parent for discontinues?)

isaacvetter:
But, as an array, `toBeDiscontinued` doesn't enable pre-fetches. 

Probably, each order being discontinued should be a full FHIR resource. A CDS Service givign advice on discontinued orders wil likely always want more data about the orders than just the FHIR id. Further, a typical clinical user workflow will limit the number of orders being discontinued at the same time.

PR #600, opened 2022-01-03
OO in prefetch

isaacvetter:
Fixes [FHIR-34274](https://jira.hl7.org/browse/FHIR-34274)


docs/specification/current.md
```diff
@@ -230,7 +230,7 @@ The resulting response is passed along to the CDS Service using the `prefetch`
...
+ The CDS Client MUST NOT include any prefetch template key that it chooses not to satisfy. If the CDS Client receives an OperationOutcome from the FHIR server when prefetching any data, this SHOULD be included in the prefetch to avoid the CDS Service calling the FHIR server again merely to receive the same error, and also to be consistent with the FHIR specification. However, if the CDS Client encounters an error while prefetching any data, the prefetch template key MUST NOT be included in the request to the CDS Service. If the CDS Client has no data to populate a template prefetch key, the prefetch template key MUST have a value of __null__. Note that the __null__ result is used rather than a bundle with zero entries to account for the possibility that the prefetch url is a single-resource request.
```
isaacvetter, line 233:
```suggestion
The CDS Client MUST NOT send any prefetch template key that it chooses not to satisfy. If the CDS Client encounters errors prefetching the requested data, OperationOutcome(s) SHOULD be used to communicate those errors to prevent the CDS Service from incurring an unneeded follow-up query. CDS Clients MUST omit the prefetch key if relevant details cannot be provided (e.g. intermittent connectivity issues). CDS Services SHOULD check any prefetched data for the existence of OperationOutcomes. If the CDS Client has no data to populate a template prefetch key, the prefetch template key MUST have a value of __null__. Note that the __null__ result is used rather than a bundle with zero entries to account for the possibility that the prefetch url is a single-resource request.
```
Per Dennis' add'l feedback provided during jira rec: https://jira.hl7.org/browse/FHIR-34274

isaacvetter:
I think I got this right, but this PR could use a detailed review to ensure we're faithfully implementing the intent of: https://jira.hl7.org/browse/FHIR-34274.

brynrhodes (approved):

PR #547, opened 2020-11-11
J#28769 - improve feedback examples

isaacvetter:
Resolves FHIR-28769

PR #602, opened 2022-01-03
more realistic timestamp

isaacvetter:
Fixes https://jira.hl7.org/browse/FHIR-28739


isaacvetter:
I'm considering this a typo fix and not getting review. Pre-applying.

dennispatterson:
I would consider this jira resolved with https://github.com/cds-hooks/docs/pull/547

dennispatterson (approved):
Since this was already resolved per previous comment, I would consider the PR optional.  Nevertheless, here's approval if we still want to define a time other than midnight UTC.

PR #603, opened 2022-01-03
reference card.indicator

isaacvetter:
Fixes https://jira.hl7.org/browse/FHIR-28714

### DIFF
--- a/docs/specification/current.md
+++ b/docs/specification/current.md
@@ -467,7 +467,7 @@ For successful responses, CDS Services SHALL respond with a 200 HTTP response wi
 
 Each card contains decision support from the CDS Service. Generally speaking, cards are intended for display to an end user. The data format of a card defines a very minimal set of required attributes with several more optional attributes to suit a variety of use cases. For instance, narrative informational decision support, actionable suggestions to modify data, and links to SMART apps.
 
-> Note that because the CDS client may be invoking multiple services from the same hook, there may be multiple responses related to the same information. This specification does not address these scenarios specifically; both CDS Services and CDS Clients should consider the implications of multiple CDS Services in their integrations.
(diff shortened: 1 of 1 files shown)


docs/specification/current.md
```diff
@@ -467,7 +467,7 @@ For successful responses, CDS Services SHALL respond with a 200 HTTP response wi
 
 Each card contains decision support from the CDS Service. Generally speaking, cards are intended for display to an end user. The data format of a card defines a very minimal set of required attributes with several more optional attributes to suit a variety of use cases. For instance, narrative informational decision support, actionable suggestions to modify data, and links to SMART apps.
 
-> Note that because the CDS client may be invoking multiple services from the same hook, there may be multiple responses related to the same information. This specification does not address these scenarios specifically; both CDS Services and CDS Clients should consider the implications of multiple CDS Services in their integrations.
+> Note that because the CDS client may be invoking multiple services from the same hook, there may be multiple responses related to the same information. This specification does not address these scenarios specifically; both CDS Services and CDS Clients should consider the implications of multiple CDS Services in their integrations and are invited to consider [card.indicator](#card-attributes).
```
dennispatterson, line 470: I think this should be kept generic.  Topic or other future metadata could influence the CDS Client, as well.  As soon as one attribute is listed, the ask will come in to list another. 

```suggestion
> Note that because the CDS client may be invoking multiple services from the same hook, there may be multiple responses related to the same information. This specification does not address these scenarios specifically; both CDS Services and CDS Clients should consider the implications of multiple CDS Services in their integrations and are invited to consider [card attributes](#card-attributes) when determining prioritization and presentation options.
```

dennispatterson (approved):

PR #604, opened 2022-01-05
only authenticate when required

isaacvetter:
Fixes https://jira.hl7.org/browse/FHIR-28694

I'm not sure that I like this.


docs/specification/current.md
```diff
@@ -864,10 +864,10 @@ The authorization server is responsible for enforcing restrictions on the CDS Se
 
 The service agreement negotiated between the CDS Client vendor/provider and the CDS Service provider will include obligations the CDS Client vendor/provider commits to the CDS Service provider. Some agreements MAY include the use of mutual TLS, in which both ends of the channel are authenticated.
 
-However, mutual TLS is impractical for many organizations. In the absence of mutual TLS, only the CDS Service endpoint will be authenticated because the CDS Client initiates the TLS channel set-up.  To enable the CDS Service to authenticate the identity of the CDS Client, CDS Hooks uses digitally signed [JSON web tokens (JWT)](https://jwt.io/) ([rfc7519](https://tools.ietf.org/html/rfc7519)).
+However, mutual TLS is impractical for many organizations. In the absence of mutual TLS, only the CDS Service endpoint will be authenticated because the CDS Client initiates the TLS channel set-up.  To enable the CDS Service to authenticate the identity of the CDS Client, CDS Hooks uses digitally signed [JSON web tokens (JWT)](https://jwt.io/) ([rfc7519](https://tools.ietf.org/html/rfc7519)). CDS Services SHOULD require authentication.
```
jmandel, line 867: >CDS Services SHOULD require authentication.

This isn't reflected in the issue resolution and I don't agree with the claim in general. Open source immunization decision support system that hosts a public website and a public CDS Hooks service... seems like a core use case and good public benefit.

jmandel, line 867:
```suggestion
However, mutual TLS is impractical for many organizations. In the absence of mutual TLS, only the CDS Service endpoint will be authenticated because the CDS Client initiates the TLS channel set-up.  To enable the CDS Service to authenticate the identity of the CDS Client, CDS Hooks uses digitally signed [JSON web tokens (JWT)](https://jwt.io/) ([rfc7519](https://tools.ietf.org/html/rfc7519)). CDS Services SHOULD require authentication if invoking the service poses any risk of exposing sensitive data to the caller.
```

This blanket "SHOULD" isn't reflected in the issue resolution and I don't agree with the claim in general. Depends on your service and whether calling the service poses any threat and leaking sensitive data. Open source immunization decision support system that hosts a public website and a public CDS Hooks service... seems like a core use case and good public benefit. Can we relax it with the clause I've proposed here?

jmandel (changes requested):

jmandel (approved):

PR #607, opened 2022-01-25
fix mispelling

isaacvetter:
fixes: https://jira.hl7.org/browse/FHIR-28652


isaacvetter:
These are all typos and mispellings. So not asking for additional review.

PR #610, opened 2022-01-26
actions are transactions

### LINKED ISSUE #419: Transaction behavior on suggestions
When returning suggestions back from a decision support service, it's possible that the server can return multiple "create" actions as part of a single suggestion - and that the resources to be created can have references to each other.  As such, the creates (or updates) will need to be processed in a transactional manner.   At present, the CDS Hook specification does not provide any guidance setting this expectation, nor does it provide guidance about how resources are to point to each other.  Suggest adding guidance that states that resource ids must be included in the resource instances to be "created" if there are interdependency issues and that the actions will be processed as a FHIR "transaction" with the full URLs being inferred as the base URL of the CDS client's server.

isaacvetter:
Fixes FHIR-35878
Fixes #419

### DIFF
--- a/docs/specification/current.md
+++ b/docs/specification/current.md
@@ -548,7 +548,9 @@ Field | Optionality | Type | Description
 `label` | REQUIRED | *string* | Human-readable label to display for this suggestion (e.g. the CDS Client might render this as the text on a button tied to this suggestion).
 `uuid` | OPTIONAL | *string* | Unique identifier, used for auditing and logging suggestions.
 `isRecommended` | OPTIONAL | *boolean* | When there are multiple suggestions, allows a service to indicate that a specific suggestion is recommended from all the available suggestions on the card. CDS Hooks clients may choose to influence their UI based on this value, such as pre-selecting, or highlighting recommended suggestions. Multiple suggestions MAY be recommended, if `card.selectionBehavior` is `any`.
-`actions` | OPTIONAL | *array* of **[Actions](#action)** | Array of objects, each defining a suggested action. Within a suggestion, all actions are logically AND'd together, such that a user selecting a suggestion selects all of the actions within it.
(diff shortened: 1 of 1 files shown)


docs/specification/current.md
```diff
@@ -548,7 +548,9 @@ Field | Optionality | Type | Description
...
-`actions` | OPTIONAL | *array* of **[Actions](#action)** | Array of objects, each defining a suggested action. Within a suggestion, all actions are logically AND'd together, such that a user selecting a suggestion selects all of the actions within it.
+`actions` | OPTIONAL | *array* of **[Actions](#action)** | Array of objects, each defining a suggested action. Within a suggestion, all actions are logically AND'd together, such that a user selecting a suggestion selects all of the actions within it. When a suggestion contains multiple actions, the actions SHOULD be processed as per FHIR's rules for processing [transactions](https://hl7.org/fhir/http.html#trules) with the CDS Client's `fhirServer` as the base url for the inferred full URL of the transaction bundle entries. (Specifically, creates happen first, then updates, then deletes).
```
dennispatterson, line 551: Align example with FHIR's rules
```suggestion
`actions` | OPTIONAL | *array* of **[Actions](#action)** | Array of objects, each defining a suggested action. Within a suggestion, all actions are logically AND'd together, such that a user selecting a suggestion selects all of the actions within it. When a suggestion contains multiple actions, the actions SHOULD be processed as per FHIR's rules for processing [transactions](https://hl7.org/fhir/http.html#trules) with the CDS Client's `fhirServer` as the base url for the inferred full URL of the transaction bundle entries. (Specifically, deletes happen first, then creates, then updates).
```

brynrhodes (approved):

lmckenzi:
Agree with the alignment

Issue #377, opened 2018-08-01
How to support pre-fetch tokens for objects in context?


isaacvetter:
As part of the [specification of each hook](https://cds-hooks.org/hooks/template/#context), each field defined in `context` is explicitly labeled whether it can be used as a Prefetch Token.

Prefetch tokens are important because they are how a CDS service specifies any specific FHIR queries to be run as part of its pre-fetch during the request. A CDS service's discovery endpoint specifies it's pre-fetch as a [pre-fetch template](https://cds-hooks.org/specification/1.0/#prefetch-template).

Currently, our pre-fetch token syntax is nice and simple - it only supports top-level elements. For example, a service responding to the `[order-review](https://cds-hooks.org/hooks/order-review/)` hook can request pre-fetch data about the patient, like so:
```
{
  "prefetch": {
    "p": "Patient/{{context.patientId}}"
  }
}
```

But the service has no way to request pre-fetch data that's based upon the `orders` object which is also sent in `context`. Unlike the simple, valued `{{context.patientId}}`, relating a prefetch template to the orders being sent would necessitate a prefetch key such as: `{{context.orders.MedicationRequest.id[0]}}` or something even more complicated using FHIRPath. 

This inability to refer to objects in `context` is a gap.

Is there a simple, targeted method to enable this? What about something as straighforward as _context.<prefetch key>.<FHIR resource name>.id_ returning a simple comma-delimited string of FHIR identifiers that could be subsequently used in a search?

lmckenzi:
What Isaac actually proposed where it says _context...id_ before Github intervened was `context.<prefetch key>.<FHIR resource name>.id`

I'd be in favor of this.  The reason for putting resource names in is to allow you to search by each resource because the ids aren't unique across resource types.  The comma-separated approach allows you to say something like DeviceRequest?_id={{context.orders.medicationRequest.id}}&_include=DeviceRequest.device

Though that does mean that your prefetch space is going to contain both device requests and devices, the former being redundant with what's already in the hook invocation.

kpshek:
> I'd be in favor of this. The reason for putting resource names in is to allow you to search by each resource because the ids aren't unique across resource types.

So is `context.<prefetch key>.<FHIR resource name>.id` only important in the case where the context value can multiple FHIR resource types?

In the hooks I've seen prototyped today, if the context field value relates to a FHIR resource, there is just a single resource in use so using `context.<prefetch key>.id` as the prefetch token is not ambiguous.

[...]

PR #612, opened 2022-02-07
Added support for enhanced prefetch - and clarified prefetch expectat…

lmckenzi:
…ions in general.

Specifically intended to address #377 and provide support for referencing information related to data in "in-memory" orders

### DIFF
--- a/docs/specification/current.md
+++ b/docs/specification/current.md
@@ -192,7 +192,7 @@ Similarly, each CDS Client will decide what FHIR resources to authorize and to p
 
 ### Prefetch Template
 
-A _prefetch template_ is a FHIR [`read`](http://hl7.org/fhir/http.html#read) or [`search`](http://hl7.org/fhir/http.html#search) request that describes relevant data needed by the CDS Service. For example, the following is a prefetch template for hemoglobin A1c observations:
+A _prefetch template_ is a FHIR [`read`](http://hl7.org/fhir/http.html#read), collection of FHIR [`read`](http://hl7.org/fhir/http.html#read) calls, or a [`search`](http://hl7.org/fhir/http.html#search) request that describes relevant data needed by the CDS Service. For example, the following is a prefetch template for hemoglobin A1c observations:
 
 ```
 Observation?patient={{context.patientId}}&code=4548-4&_count=1&sort:desc=date
@@ -220,7 +220,7 @@ A CDS Client MAY choose to honor some or all of the desired prefetch templates, 
 
 The CDS Client SHALL deny access to the requested resource if the user in context is not authorized to access the resource.
 
-As part of preparing the request, a CDS Client processes each prefetch template it intends to satisfy by replacing the prefetch tokens in the prefetch template to construct a relative FHIR request URL. This specification is not prescriptive about how this request is actually processed. The relative URL may be appended to the base URL for the CDS Client's FHIR server and directly invoked, or the CDS Client may use internal infrastructure to satisfy the request in the same way that invoking against the FHIR server would.
+As part of preparing the request, a CDS Client processes each prefetch template it intends to satisfy by replacing the prefetch tokens in the prefetch template and manipulating the results to construct RESTful call. This specification is not prescriptive about how this request is actually processed. The content may be executed as a direct GET or POST against the CDS Client's FHIR server and directly invoked, or the CDS Client may use internal infrastructure to satisfy the request in the same way that invoking against the FHIR server would.
 
 Regardless of how the CDS Client satisfies the prefetch templates (if at all), the prefetched data given to the CDS Service MUST be equivalent to the data the CDS Service would receive if it were making its own call to the CDS Client's FHIR server using the parameterized prefetch template.
 
@@ -230,7 +230,7 @@ The resulting response is passed along to the CDS Service using the `prefetch` 
 
 > Note that the reason prefetch results are not allowed to include next page links is that if the prefetched data contains only the first page of a multi-page search result, the CDS Service has no means to retrieve the subsequent pages of data. Consider, for example, a CDS Hooks implementation that does not expose a FHIR server.
 
- The CDS Client MUST NOT include any prefetch template key that it chooses not to satisfy. Similarly, if the CDS Client encounters an error while prefetching any data, the prefetch template key MUST NOT be included in the request to the CDS Service. If the CDS Client has no data to populate a template prefetch key, the prefetch template key MUST have a value of __null__. Note that the __null__ result is used rather than a bundle with zero entries to account for the possibility that the prefetch url is a single-resource request.
+The CDS Client MUST NOT include any prefetch template key that it chooses not to satisfy. Similarly, if the CDS Client encounters an error while prefetching any data, the prefetch template key MUST NOT be included in the request to the CDS Service. If the CDS Client has no data to populate a template prefetch key, the prefetch template key MUST have a value of __null__. Note that the __null__ result is used rather than a bundle with zero entries to account for the possibility that the prefetch url is a single-resource request.
 
 It is the CDS Service's responsibility to check prefetched data against its template to determine what requests were satisfied (if any) and to programmatically retrieve any additional necessary data. If the CDS Service is unable to obtain required data because it cannot access the FHIR server and the request did not contain the necessary prefetch keys, the service SHALL respond with an HTTP 412 Precondition Failed status code.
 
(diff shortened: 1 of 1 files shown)


docs/specification/current.md
```diff
@@ -240,45 +240,66 @@ A prefetch token is a placeholder in a prefetch template that is *_replaced by i
...
+* If a collection element is a Coding or Identifier, the element is converted to [token](http://hl7.org/fhir/search.html#token) syntax with code/value followed by '|' followed by system.
+5. Content expanded from a prefetch token may require HTTP escaping to ensure a valid URL fragment
+6. When dealing with prefetch tokens that resolve to relative references, in some cases, the search needs the resolved token to be an id or collection of ids rather than references. In this circumstance, the FHIRPath can be followed with ":" and the name of the resource whose references should be filtered.  Only relative references for the specified resource type are returned.  For example `{{context.draftOrders.ofType(MedicationRequest).requester.reference:Practitioner}}` would resolve to the a list of ids who are requesters that are also Practitioners.  For legacy reasons, this can also be done with the userId element by inserting the resource name into the context name.  E.g. `{{context.userPatientRoleId}}`, `{{context.userPractionerId}}`, `{{context.userPractitionerRoleId}}`, and `{{context.userRelatedPersonId}}`.
```
jmandel, line 275: I don't agree with the description of legacy equivalents, or else I don't understand it correctly. Why would the user context for invoking a hook necessarily be equivalent to the person writing the draft orders? ( I can see that they might be the same and often would be... But that's not the same as definitional equivalence)

lmckenzi: The equivalence is that in `userId`, we provide a mechanism to take a userId context that would normally resolve to "Patient/123" or "Practitioner/456, etc. and instead to reference the context element as `userPatientId` and have it resolve to exactly "123", but only if the reference was a Patient.
The new mechanism of throwing `:Patient` on the end of the path is more generic.  It could work for userId (e.g. `context.userId:Patient`), but it can also work for any other path (e.g. `context.order:ServiceRequest` or `context.draftOrders.requester.reference:Practitioner`).  

The need that we solved as a one-off with userId turns out to be a general-purpose issue.  And it's an issue that can't be solved by just changing the name of a context variable.  The new approach will work for any path that corresponds to a 'Reference.reference' or to some other string that happens to be a relative reference.

jmandel: That all makes sense, in terms of the semantics of identifying a reference. The thing I was concerned about was the expectation (implication) that the person requesting a draft order would necessarily be the same as the user.

lmckenzi: Ok.  That wasn't an intended implication at all.  If you have suggestions on how to re-word to avoid that implication, I'm happy to adjust.

jmandel: Just remove the sentence station with

>For legacy reasons, this can also be done with the userId element by inserting the resource name

That capability is to find elsewhere, right? While it's analogous it's not quite the same thing as traversing a resource path in the manner you're documenting for the new capability.

lmckenzi: No - it's no longer documented elsewhere (or it shouldn't be - if I left it in, it was by accident).  The intention is to move all of the 'search' stuff into one place.

jmandel: Okay, then I'd document it separately; confusing to try to explain it here in the context of another syntax (and it's not like you can communicate "hook user" as part of navigating fhirpath on a context resource because there's no context variable representing the user in, just a set representing user IDs)

lmckenzi: The requirement is identical in both places.  And if we had adopted this new mechanism initially, there never would have *been* a userPatientId, etc. syntax.  That syntax is now redundant and unnecessary.  So it certainly seems to be appropriate to describe it in the same place as the more general syntax that's solving the same problem...

jmandel: Maybe I'm just missing what you have in mind. Can you explain what this would have looked like if we had introduced this new capability previously? What resource would we be drilling into with fhirpath expressions? An example might clarify for me.

lmckenzi: Using the new syntax, instead of `{{context.userPatientId}}`, you'd have `{{context.userId:Patient}}` - and there'd be no such thing as userPatientId.

lmckenzi: (Obviously, given that userPatientId is in use, I'm not proposing to eliminate it.  It would remain as a legacy syntax.)

docs/specification/current.md
```diff
@@ -240,45 +240,66 @@ A prefetch token is a placeholder in a prefetch template that is *_replaced by i
...
   "prefetch": {
-    "user": "{{context.userId}}"
+    "patient": "Patient/{{context.patient}}",
+    "user": "{{context.user}}"
   }
 }
 ```
-or though a FHIR search:
+
+`{{context.patient}}` resolves to a simple id.  `{{context.user}}` resolves to a resource type and simple id separated by a '/'.  In both cases, the result of resolving the prefetch template is a single string that matches a relative URL to a single resource instance.  That relative URL could be appended to the CDS Client's server base URL to form the URL for a valid read.
```
jmandel, line 259:
```suggestion
`{{context.patient}}` resolves to a simple id (e.g., `123` for the Patient with id 123).  `{{context.user}}` resolves to a resource type and simple id separated by a '/' (e.g., `Patient/123` or `Practitioner/456`).  In both cases, the result of resolving the prefetch template is a single string that matches a relative URL to a single resource instance.  That relative URL could be appended to the CDS Client's server base URL to form the URL for a valid read.
```

Adding examples would help, but I'm still not clear what the *assumptions* here. Is this paragraph **defining** a new `context.patient` variable in context, and a new `context.user` variable?

docs/specification/current.md
jmandel, line 275: This "extra colon at the end" syntax makes it harder to see what is the "thing that needs to be parsed as FHIRPath". I'd rather see (in order of preference):

* Define "simple FHIRpath + `replaceMatches`" so developers can build a list of IDs by taking final path segments
* Define a distinct templating syntax `{{expr}:Patient}` or somesuch so the fhirpath is unambiguous
* Let CDS Clients figure out how to filter based on context (i.e., "please do what I mean" -- if searching for )
* Use existing fhirpath comment syntax like `/*Patient*/` instead of tacking on `:`. Aim would be to leverage existing parsers. If you just split on `:` you hit complications in places where the gramamr (https://github.com/HL7/FHIRPath/blob/master/spec/fhirpath.g4) allows colons.

jmandel, line 275: > For legacy reasons, this can also be done

The "legacy" examples are about "context.user*" but the other equivalent examples are about "context.draftOrders...". I'm unclear on how to translate (say) `context.userPatientRoleId` in the new syntax.

docs/specification/current.md
```diff
@@ -240,45 +240,66 @@ A prefetch token is a placeholder in a prefetch template that is *_replaced by i
...
+* The 'presciberPR' prefetch shows converting relative references to a subset of ids.
+* The 'userPR' prefetch shows the ability to filter to a subset of ids using the special legacy 'userId' naming mechanism
 
+Note:
+* Authors SHOULD NOT create prefetches that presume availability of context information not expected to be available in persisted form.  For example `MedicationRequest?_id={{context.draftOrders.id}}` would be problematic because the the draft orders aren't expected to have been persisted yet (though they might be on some clients).
 
-#### Prefetch query restrictions
+###### Prefetch query restrictions
 
 To reduce the implementation burden on CDS Clients that support CDS Services, this specification RECOMMENDS that prefetch queries only use a subset of the full functionality available in the FHIR specification. When using this subset, valid prefetch templates MUST only make use of:
```
jmandel, line 301: What is the name of this subset (so servers can say whether they are limited to it)?

docs/specification/current.md
```diff
@@ -240,45 +240,66 @@ A prefetch token is a placeholder in a prefetch template that is *_replaced by i
...
 
-* _instance_ level [read](https://www.hl7.org/fhir/http.html#read) interactions (for resources with known ids such as `Patient`, `Practitioner`, or `Encounter`)
 * _type_ level [search](https://www.hl7.org/fhir/http.html#search) interactions; e.g. `patient={{context.patientId}}`
 * Resource references (e.g. `patient={{context.patientId}}`)
```
jmandel, line 304: This prohibits chaining, yes? It'd be good to annotate the examples above to say whether they are "compatible with the restricted subset".

hstrasbe:
As we discussed at today's CDS WG meeting, we need to be able to support a request for something like all BP Observations from the last 10 days.  

As currently proposed we could retrieve the last 10 BP Observations using:
"bloodPressures": "Observation?patient={{context.patient}}&code=85354-9|http%xx/loinc.org&_sort=-date&_count=10"

However, there does not seem to be a way to request all BP Observations from the last 10 days using a prefetch template.

lmckenzi:
One option would be to use [x-fhir-query](https://build.fhir.org/fhir-xquery.html), though that's a bigger lift than what's proposed here.  We could possibly also subset x-fhir-query?

docs/specification/current.md
```diff
@@ -240,45 +240,66 @@ A prefetch token is a placeholder in a prefetch template that is *_replaced by i
 
 Prefetch tokens MUST be delimited by `{{` and `}}`, and MUST contain only the qualified path to a hook context field *_or one of the following user identifiers: `userPractitionerId`, 'userPractitioneRoleId', `userPatientId`, or `userRelatedPersonId`_*.
 
-Individual hooks specify which of their `context` fields can be used as prefetch tokens. Only root-level fields with a primitive value within the `context` object are eligible to be used as prefetch tokens. For example, `{{context.medication.id}}` is not a valid prefetch token because it attempts to access the `id` field of the `medication` field.
+Individual hooks specify which of their `context` fields can be used as prefetch tokens. Prefetch templates must start with one of these context elements and can then navigate to information available within the context element using [simple FHIRPath](https://hl7.org/fhir/fhirpath.html#simple).  For example,
```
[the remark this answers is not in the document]
lmckenzi: Fixed

docs/specification/current.md
lmckenzi: Accepted

PR #623, opened 2022-05-25
actions are transactions, recreated

isaacvetter:
Erg. Recreating #610. Awaiting Lloyd's feedback for ... 2.5 weeks (week following DevDays, 22), at which point, I'll merge. 

Fixes FHIR-35878
Fixes https://github.com/cds-hooks/docs/issues/419


lmckenzi (approved):
Agree with the correction.  Deletes do indeed happen first.

PR #632, opened 2023-01-24
Propose new questionnaire-inprogress hook

titicaca37:
We  @PTschandl and @Titicaca37 extended an existing dermatological CDS system to find similar skin diseases to comply with the CDS Hooks specification based on FHIR Questionnaire and QuestionnaireResponse resources. For our use case we specified and implemented an additional hook called questionnaire-inprogress which is to be invoked whenever a user is in the process of answering a questionnaire.

### DIFF
--- a/docs/hooks/questionnaire-inprogress.md
+++ b/docs/hooks/questionnaire-inprogress.md
@@ -1,0 +1,123 @@
+# `questionnaire-inprogress`
+
+!!! info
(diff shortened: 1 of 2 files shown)


vadi2 (commented):
This seems to step on the toes of [adaptive forms](https://hl7.org/fhir/uv/sdc/adaptive.html) in SDC, so it would be good to mention how this hook relates to that feature.

PR #634, opened 2023-02-14
rename auto-launchable to autolaunchable

isaacvetter:
per implementer feedback


vadi2:
Fine with the change, just curious as to the reason?

PR #647, opened 2023-11-09
Remove generic statement per jira FHIR-42367

rizzolol:
Removing generic statement per [jira FHIR-42367](https://jira.hl7.org/browse/FHIR-42367).


isaacvetter (approved):
Looks great, @rizzolol !

PR #656, opened 2024-03-13
Disconnect from cds-hooks.org for IG Publisher migration

isaacvetter:
Delete gh action.

Anything else needed here?


vadi2 (approved):
Not that I can think of.

@jmandel ?

jmandel:
This seems fine for a start. Eventually do we want to keep the latest build visible at cds-hooks.org, perhaps with a new GitHub action that will run the publisher and use a deploy step?

vadi2:
I can put that together. How did current deploy to cds-hooks.org get the new content over?

isaacvetter:
Hey Guys,

_>do we want to keep the latest build visible at cds-hooks.org, perhaps with a new GitHub action that will run the publisher and use a deploy step?_

The current plan is to shift cds-hooks.org to contain informative content with a focus on community engagement, with links to the published spec(s) and the IG ci build. We haven't worked out sufficient technical details yet, but mkdocs and gh actions and gh hosting work pretty well for website editing and maintenance, so one repo for the IG and another for the cds-hooks.org website? 

HL7 will also expect an IG to live in a repo within the HL7 organization. We do already have this one -- https://github.com/HL7/cds-hooks-hl7-site. So what probably makes sense is to host the IG in either HL7/cds-hooks-hl7-site or a new one, like HL7/cds-hooks and keep cds-hooks/docs as the source and publishing pipeline for cds-hooks.org.

What do you guys think about that?

vadi2:
That makes sense to me 👍

